PR #4, opened 2017-06-14
Fixed fname problem in get_files

deeptavker:


prabhuramachandran:
Can you please add tests for this?

pysph/solver/tests/test_solver_utils.py
```diff
@@ -12,7 +12,39 @@
...
+        self.dirname = join(self.root, self.fname + '_output')
+        os.mkdir(self.dirname)
+        self.files = [
+            join(
+                self.dirname,
+                self.fname+'_'+str(i)+'.npz'
+            )
+            for i in range(11)
+        ]
+        for names in self.files:
```
prabhuramachandran: `for name in self.files` is better.

pysph/solver/tests/test_solver_utils.py
```diff
@@ -12,7 +12,39 @@
...
+            open(names, 'a').close()
```
prabhuramachandran: Might be better to do:

```
    with open(name, 'w') as fp:
        fp.write('')
```

PR #6, opened 2017-06-15
Measure coverage

deeptavker:
Added .coveragerc and modified setup.py so that when the environment variable `COVERAGE` is set to `true` linetracing is enabled and coverage for cython code can be measured. 

Pytest results : 

For COVERAGE=true
= 4 failed, 411 passed, 9 skipped, 44 deselected in 3186.40 seconds =

For COVERAGE != true 
= 415 passed, 9 skipped, 44 deselected in 242.63 seconds =


setup.py
```diff
@@ -23,6 +23,16 @@ def check_output(*popenargs, **kwargs):
 
 import sys
 
+if os.environ.get('COVERAGE','') == 'true':
```
prabhuramachandran: Can this just be any non-empty string?  setting it to 'true' exactly is too easy to get wrong.

setup.py
```diff
@@ -649,7 +680,8 @@ def setup_package():
         ext_modules = cythonize(
             ext_modules, compile_time_env=compile_env,
             include_path=list(include_path),
-            language="c++"
+            language="c++",
+            compiler_directives=COMPILER_DIRECTIVES,
```
prabhuramachandran: `cythonize` is only called for Cython >= 0.25, will the tracing still work in those cases?

setup.py
prabhuramachandran: Also can you check if COVERAGE is already used elsewhere by the `coverage` module?

deeptavker: Updated, the environment variable can be any non-empty string. 

No, sir, `COVERAGE` is not used anywhere by the `coverage` module. But I think it would be better if we change it to - `MEASURE_COVERAGE`. Shall I change it?

setup.py
deeptavker: No, it won't. Do we need to add support for Cython < 0.25?

prabhuramachandran: Yes, that would be good if possible.

setup.py
prabhuramachandran: `MEASURE_COVERAGE` is a tad long.  Maybe just `COVERAGE` is good for now.

setup.py
```diff
@@ -23,6 +23,16 @@ def check_output(*popenargs, **kwargs):
 
 import sys
 
+if len(os.environ.get('COVERAGE','')) > 0:
+    MACROS = [("CYTHON_TRACE","1"),("CYTHON_TRACE_NOGIL","1")]
```
prabhuramachandran: Please follow PEP8 style, spaces after commas.

setup.py
```diff
@@ -23,6 +23,16 @@ def check_output(*popenargs, **kwargs):
...
+    COMPILER_DIRECTIVES = {"linetrace":True}
```
prabhuramachandran: Same here for the dictionary.

prabhuramachandran:
Thanks! Merging.

PR #8, opened 2017-06-15
Float to int division

quang-ha:
Fix a bug float to int that failed cause failure in dam_break_3d before.


prabhuramachandran:
Thanks, looks good to me.

PR #10, opened 2017-07-09
Jupyter Notebook Viewer 2D 

deeptavker:
3D still needs some work, 2D is done with the additional options of particle size, vector scale and save as pdf. 

ui

### DIFF
--- a/pysph/tools/ipython.py
+++ b/pysph/tools/ipython.py
@@ -193,18 +193,9 @@ class Viewer(object):
         self.show_log()
 
 
-
-
-
-
 class Viewer2D(Viewer):
 
     def _create_widgets(self):
-        '''
-        For handlers to work,
-        make sure description of each widget starts like fluid_widname.
-
-        '''
 
         from pysph.solver.utils import load
         import ipywidgets as widgets
@@ -239,10 +230,13 @@ class Viewer2D(Viewer):
                     'None'
                 ]+temp_data[particles_type].output_property_arrays,
                 value='rho',
-                description=particles_type + "_scalar",
+                description="scalar",
                 disabled=False,
                 layout=widgets.Layout(width='240px', display='flex')
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['scalar'].owner = particles_type
             self._widgets['particles'][particles_type]['scalar'].observe(
                 self._scalar_handler,
                 'value',
@@ -251,11 +245,15 @@ class Viewer2D(Viewer):
                 'particles'
             ][particles_type]['legend'] = widgets.ToggleButton(
                 value=False,
-                description=particles_type + "_legend",
+                description="legend",
                 disabled=False,
                 tooltip='Description',
-                layout=widgets.Layout(width='240px', display='flex')
+                layout=widgets.Layout(width='80px', display='flex')
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['legend'].owner = particles_type
+
             self._widgets['particles'][particles_type]['legend'].observe(
                 self._legend_handler,
                 'value',
@@ -266,10 +264,13 @@ class Viewer2D(Viewer):
             ][particles_type]['vector'] = widgets.Text(
                 value='',
                 placeholder='variable1,variable2',
-                description=particles_type + '_vector',
+                description='vector',
                 disabled=False,
                 layout=widgets.Layout(width='240px', display='flex')
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['vector'].owner = particles_type
             self._widgets['particles'][particles_type]['vector'].observe(
                 self._vector_handler,
                 'value',
@@ -281,9 +282,12 @@ class Viewer2D(Viewer):
                 max=100,
                 step=1,
                 value=25,
-                description=particles_type + '_vector_width',
+                description='vector width',
                 layout=widgets.Layout(width='300px'),
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['vector_width'].owner = particles_type
 
             self._widgets[
                 'particles'
@@ -298,9 +302,12 @@ class Viewer2D(Viewer):
                 max=100,
                 step=1,
                 value=55,
-                description=particles_type + '_vector_scale',
+                description='vector scale',
                 layout=widgets.Layout(width='300px'),
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['vector_scale'].owner = particles_type
             self._widgets[
                 'particles'
             ][particles_type]['vector_scale'].observe(
@@ -314,9 +321,12 @@ class Viewer2D(Viewer):
                 max=50,
                 step=1,
                 value=10,
-                description=particles_type + '_scalar_size',
+                description='scalar size',
                 layout=widgets.Layout(width='300px'),
             )
+            self._widgets[
+                'particles'
+            ][particles_type]['scalar_size'].owner = particles_type
             self._widgets[
                 'particles'
             ][particles_type]['scalar_size'].observe(
@@ -337,13 +347,14 @@ class Viewer2D(Viewer):
                 VBox([
                     Label(particles_type),
                     self._widgets['particles'][particles_type]['scalar'],
-                    self._widgets['particles'][particles_type]['legend'],
                     self._widgets['particles'][particles_type]['vector'],
                     self._widgets[
                         'particles'
                     ][particles_type]['vector_scale'],
                     self._widgets['particles'][particles_type]['vector_width'],
                     self._widgets['particles'][particles_type]['scalar_size'],
+                    self._widgets['particles'][particles_type]['legend'],
+
                 ],
                     layout=Layout(display='flex'))
             )
@@ -357,6 +368,7 @@ class Viewer2D(Viewer):
                 ]
             )
         )
+
     def _configure_plot(self):
         '''
         Set attributes for plotting.
@@ -476,7 +488,7 @@ class Viewer2D(Viewer):
 
         temp_data = self.get_buffer(self._widgets['frame'].value)['arrays']
 
-        particles_type = change['owner'].description.split('_')[0]
+        particles_type = change['owner'].owner
 
         if particles_type in self._scatters.keys():
             if self._scatters[particles_type] in self._scatter_ax.collections:
@@ -509,7 +521,7 @@ class Viewer2D(Viewer):
             self._widgets['frame'].value
         )['arrays']
 
-        particles_type = change['owner'].description.split('_')[0]
+        particles_type = change['owner'].owner
 
         if particles_type in self._vectors.keys():
             if self._vectors[particles_type] in self._scatter_ax.collections:
@@ -550,7 +562,7 @@ class Viewer2D(Viewer):
     def _vector_scale_handler(self, change):
         from IPython.display import display, clear_output
         # the widget value must already have updated.
-        particles_type = change['owner'].description.split('_')[0]
+        particles_type = change['owner'].owner
         if particles_type in self._vectors.keys():
             self._vectors[particles_type].scale = change['new']
         clear_output(wait=True)
@@ -558,7 +570,7 @@ class Viewer2D(Viewer):
 
     def _scalar_size_handler(self, change):
         from IPython.display import display, clear_output
-        particles_type = change['owner'].description.split('_')[0]
+        particles_type = change['owner'].owner
         if particles_type in self._scatters.keys():
             self._scatters[particles_type].set_sizes([change['new']])
         clear_output(wait=True)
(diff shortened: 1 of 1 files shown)


pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+    '''A class for the pysph jupyter notebook Viewer2D.
+
+    Attributes
+    ----------
+
+        path              String containing the path of the output directory
+        paths_list        List of paths to the output files
+        cache             Cached data
+
+    Methods
```
prabhuramachandran: Do you really need the attributes and methods listed.  Instead provide an example of how to use this.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+        show_results
+        show_all
+
+    '''
+
+    def __init__(self, path, cache=True):
+
+        # Imports #
+
+        from pysph.solver.utils import get_files, load
```
prabhuramachandran: `load` is not used here.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+
+        # Caching #
+        # Note : Caching is only used by get_buffer and widget handlers.
+        if cache:
+            self.cache = {}
+        else:
+            pass
+
+    def get_buffer(self, frame):
+        '''Return particle arrays for a given frame number with caching.
```
prabhuramachandran: Why the name "buffer", instead can you not just use `get_frame`?

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+        >>> sample = Viewer2D('/home/deep/pysph/trivial_inlet_outlet_output/')
+        >>> sample.get_buffer(12)
+        {'arrays': {'fluid': <pysph.base.particle_array.ParticleArray at 0x7f3f7d144d60>,
+        'inlet': <pysph.base.particle_array.ParticleArray at 0x7f3f7d144b98>,
+        'outlet': <pysph.base.particle_array.ParticleArray at 0x7f3f7d144c30>},
+        'solver_data': {'count': 240, 'dt': 0.01, 't': 2.399999999999993}}
+
+
+        '''
+        if hasattr(self, "_cbars"):
```
prabhuramachandran: Instead define `_cbars` to None in `__init__` and check if it is None or not.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+            if frame in self.cache.keys():
```
prabhuramachandran: `if frame in self.cache`.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+                temp_data = self.cache[frame]
+            else:
+                from pysph.solver.utils import load
```
prabhuramachandran: Why can't you put the import at the top as you seem to need it anyway, it isn't optional so might as well put it at the top.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+            from pysph.solver.utils import load
+            temp_data = load(self.paths_list[frame])
+
+        return temp_data
+
+    def show_log(self):
+        '''
+        Prints the content of log file.
+        '''
+        import glob
```
prabhuramachandran: Put the import at the top.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+        print("Printing log : \n\n")
+        path = self.path + "*.log"
+        logfile = open(glob.glob(path)[0], 'r')
```
prabhuramachandran: Use a context manager for the file.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+    def show_results(self):
+        '''
+        Show if there are any png, jpeg, jpg, or bmp images.
+        '''
+
+        import glob
+        from IPython.display import Image, display
+
+        imgs = tuple()
+        for extension in ['png', 'jpg', 'jpeg', 'bmp']:
```
prabhuramachandran: Can we also show any PDFs?  Should we?  Will IPython show them?

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+        Print contents of the .info file present in the output directory,
+        keys present in results.npz,  number of files and
+        information about paricle arrays.
+
+        Examples
+        --------
+
+        >>> sample = Viewer2D('/home/deep/pysph/trivial_inlet_outlet_output/')
+        >>> sample.show_info()
+        Printing info :
```
prabhuramachandran: I think this example is not really needed here, it just makes for a long docstring without really adding too much value.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+            'm', 'h', 'pid', 'gid', 'tag']
+
+        Keys in results.npz :
+        ['t', 'x_max']
+        '''
+        # Imports #
+
+        import json
+        import glob
+        from pysph.solver.utils import load
```
prabhuramachandran: Imports to the top.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,641 @@
...
+
+class Viewer2D(Viewer):
+
+    def _create_widgets(self):
+
+        from pysph.solver.utils import load
+        import ipywidgets as widgets
+
+        temp_data = load(self.paths_list[0])['arrays']
+        self._widgets = {}
```
prabhuramachandran: Instead of adding the widgets to this dictionary why not just add them to self or create another object called Viewer2DWidgets which does all this and you create a Viewer2DWidgets.  The repetition of `self._widgets` in the code is too distracting and makes it hard to read the code.

deeptavker: I did that so that I can create a hierarchy such as widgets -> particles -> boundary -> scalar_property

I can create another object called `ViewerWidgets` but then again, it'll be like so : 
`self.widgets = ViewerWidgets(path)` 
And then again I'll have to use `self.widgets['particles'][particles_type]['scalar_Scale']` and so on.

prabhuramachandran: Well what I meant was not to have to navigate via the dictionary interface but using attributes, i.e. `self.widgets.particles['xxx'].scalar_scale` etc.  This makes it easier to write, type, auto-complete etc.

prabhuramachandran: Also before you push more commits, please update to master, otherwise the tests will keep failing.

pysph/tools/ipython.py
deeptavker: I'm trying to embed it like this but so far I can't see any output. 

```python
from IPython.display import HTML
HTML('<iframe src=file:///home/deep/Downloads/1366.pdf width=700 height=350></iframe>')
```

prabhuramachandran: If this is a pain, forget it.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+
+    def show_all(self):
+        self.show_info()
+        self.show_results()
+        self.show_log()
+
+
+class Viewer2DWidgets(object):
+
+    def __init__(self, file, file_count, handlers):
```
prabhuramachandran: Still too complex.  You don't need handlers at all, you could as well have passed the Viewer2D object and called observe on `viewer._frame_handler` etc.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+                placeholder='example.pdf',
+                description='Save figure',
+                disabled=False,
+                layout=widgets.Layout(width='240px', display='flex')
+        )
+        self.save_figure.on_submit(handlers['save_figure'])
+
+        self.particles = {}
+
+        class extend(object):
```
prabhuramachandran: This is a sign that the design is still not clean enough.  If you look at the code below, what you can do is create a class called 
```
class ParticleArrayWidgets(object):
   ...
```
which is given your temporary particle array and which uses that to construct the widgets.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+            pass
+
+        for particles_type in temp_data.keys():
+
+            self.particles[particles_type] = extend()
```
prabhuramachandran: You can make this so much easier if you do this:
```
self.particles[particles_type] = ParticleArrayWidgets(temp_data[particles_type])
```

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+    >>> sample = ipython.Viewer2D('/home/uname/pysph_files/dam_Break_2d_output')
+    >>> sample.interactive_plot()
+    >>> sample.show_log()
+    >>> sample.show_info()
+    '''
+
+    def _create_widgets(self):
+
+        handlers = {}
+        handlers['frame'] = self._frame_handler
```
prabhuramachandran: Not needed.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+        '''
+
+        from ipywidgets import HBox, VBox, Label, Layout
+
+        items = []
+        for particles_type in self._widgets.particles.keys():
+            items.append(
+                VBox([
+                    Label(particles_type),
+                    self._widgets.particles[particles_type].scalar,
```
prabhuramachandran: You seem to do this every time.  Instead of referring to `self._widgets.particles[particles_type].scalar`, reference this once above like so:
```
pa_widgets = self._widgets.particles['particles_types`]
...
VBox([
...
pa_widgets.scalar,
pa_widgets.vector,
...
]
...
)
```
However, in this case you can clean this up further as it smells that you create a VBox out of the widgets that belong to the widgets in this class.  Instead push the vbox creation also to the `ParticleArrayWidgets` into a method that simply returns a suitable VBox.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+        # _vectors #
+
+        for vct in self._vectors.values():
+            if vct in self._scatter_ax.collections:
+                self._scatter_ax.collections.remove(vct)
+
+        self._vectors = {}
+        for particles_type in self._widgets.particles.keys():
+            if self._widgets.particles[particles_type].vector.value != '':
+                self._vectors[particles_type] = self._scatter_ax.quiver(
```
prabhuramachandran: Again same pattern, repeated `temp_data[particles_type]` repeated `self._widgets.particle[particle_type]` etc.  You can make the code much easier to read and write than this.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,590 @@
...
+        particles_type = change['owner'].owner
+
+        if particles_type in self._vectors.keys():
+            if self._vectors[particles_type] in self._scatter_ax.collections:
+                self._scatter_ax.collections.remove(
+                    self._vectors[particles_type]
+                )
+
+        if change['new'] != '':
+            self._vectors[particles_type] = self._scatter_ax.quiver(
```
prabhuramachandran: Same here.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,523 @@
...
+            )
+
+    def _create_vbox(self):
+
+        from ipywidgets import HBox, VBox, Label, Layout
+        items = []
+        for particles_type in self.particles.keys():
+            pa_widgets = self.particles[particles_type]
+            items.append(
+                VBox([
```
prabhuramachandran: Can the vbox also be pushed to the `ParticleArrayWidgets`?  It smells that you are accessing the attributes of it to create a vbox here.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,523 @@
...
+        # Caching #
+        # Note : Caching is only used by get_frame and widget handlers.
+        if cache:
+            self.cache = {}
+        else:
+            self.cache = None
+
+    def get_frame(self, frame):
+        '''Return particle arrays for a given frame number with caching.
+
```
prabhuramachandran: Can you remove the extra whitespace here.

pysph/tools/ipython.py
```diff
@@ -0,0 +1,523 @@
...
+            )
+
+        return VBox(
+                [
+                    HBox(items, layout=Layout(display='flex')),
+                    self.frame,
+                    self.save_figure
+                ]
+            )
+
```
prabhuramachandran: Again PEP8, should have two newlines between classes and functions.

prabhuramachandran:
The changes look ok.  Can you just fix the remaining pep8 issues?

prabhuramachandran:
@deeptavker -- what about the documentation?  Can you add a little bit of documentation in here:
https://github.com/pypr/pysph/blob/master/docs/source/tutorial/circular_patch_simple.rst

prabhuramachandran:
I will merge this for now.  I don't like the module name either and will change it later.  Thanks!

PR #12, opened 2017-07-15
Installing in Anaconda2 using .yml file.

dineshadepu:
Dear sir,

The following procedure only works for Python 2. I checked for conda 3 but there is problem with dependencies of Mayavi. If you are OK with it python 3 could also be added.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -18,8 +18,8 @@ Depending on your chosen Python distribution, simply follow the instructions
 and links referred therein.
 
 .. contents::
-    :local:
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
Hi Dinesh, thanks for doing this.  Can you please wait a bit for this as the dependency stuff is changing a little and am hoping to have a better idea by next week.  Also please avoid sending changes with a lot of whitespace fixes as it is very hard to review such a PR.  For now can we close this and reopen after we discuss.

dineshadepu:
Will review it next week.

PR #16, opened 2017-07-31
Geometry module commits

sankasuraj:
Added arguments for number of layers for get_2d_tank and get_2d_wall functions and started test functions for geometry.


prabhuramachandran:
@sankasuraj -- please fix the broken test on travis that shows up on Python3.

pysph/tools/tests/test_geometry.py
```diff
@@ -0,0 +1,245 @@
+import numpy as np
+import unittest
+
+from pysph.tools.geometry import distance, distance_2d
```
prabhuramachandran: Instead you could have done the following:
```
import pysph.tools.geometry as G
# in your code
G.distance(...)
```
You could fix this later.

PR #28, opened 2017-08-31
SPHysics 2d

sankasuraj:


sankasuraj:
Pushed 5 out of the 7 sphysics 2d cases.
I am leaving the rigid body collisions part out of it for now and closing the PR.
Will finish the remaining 2 in a couple of days

prabhuramachandran:
@sankasuraj -- thanks!  Overall it looks good, but do we need to simulate these cases for 10 seconds each?  The tmax used in the sphysics cases is very different and this will significantly reduce the simulation time also.  For example case1 seems to have a default tf=3.0.  Case 2 has tmax of 1.2 seconds, case3, 5, case4, 3 etc.  Can you please fix this?

sankasuraj:
I know that the total time is different from that of SPHysics. If we just run the simulation for whatever they have given the simulation looks a bit incomplete so I have changed them even while running in SPHysics, 10s is a good enough time for the whole simulation. 

Alternatively I can add an user option for tf so that the user can run it for whatever time they seem fit with the default time like that in sphysics. So which one do you suggest I do?

prabhuramachandran:
There is already an option to set the total time, just keep the defaults to what sphysics has.

sankasuraj:
Updated the tf values and the total simulation time

PR #30, opened 2017-09-03
Added 3d Jupyter Notebook Viewer

deeptavker:
The 3D viewer currently supports:

1) ipyvolume scatter (any output property) and quiver plots (for <u,v,w> )
2) Changeable colormaps for scatter plots and corresponding legends
3) Changeable glyph sizes for both scatter and vector plots 
----------

ui3d620
uixyz
ui3d3dui

--------------

### DIFF
--- a/pysph/tools/ipy_viewer.py
+++ b/pysph/tools/ipy_viewer.py
@@ -146,18 +146,18 @@ class Viewer(object):
 
 class ParticleArrayWidgets(object):
 
-    def __init__(self, particlearray, particles_type):
-        self.particles_type = particles_type
+    def __init__(self, particlearray):
+        self.array_name = particlearray.name
         self.scalar = widgets.Dropdown(
             options=[
                 'None'
-            ]+particlearray.output_property_arrays,
+            ] + particlearray.output_property_arrays,
             value='rho',
             description="scalar",
             disabled=False,
             layout=widgets.Layout(width='240px', display='flex')
         )
-        self.scalar.owner = particles_type
+        self.scalar.owner = self.array_name
         self.scalar_cmap = widgets.Dropdown(
             options=list(map(str, plt.colormaps())),
             value='viridis',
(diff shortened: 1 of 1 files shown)


pysph/tools/ipy_viewer.py
```diff
@@ -540,3 +541,294 @@ def _save_figure_handler(self, change):
...
+
+
+class ParticleArrayWidgets3D(object):
+
+    def __init__(self, particlearray, particles_type):
+        self.particles_type = particles_type
+        self.scalar = widgets.Dropdown(
+            options=[
+                'None'
+            ]+particlearray.output_property_arrays,
```
prabhuramachandran: PEP8, space around +

pysph/tools/ipy_viewer.py
```diff
@@ -540,3 +541,294 @@ def _save_figure_handler(self, change):
...
+            pa_widgets.scalar_cmap.observe(self._scalar_cmap_handler, 'value')
+
+    def interactive_plot(self):
+        self._create_widgets()
+        self.scatters = {}
+        display(self._widgets._create_vbox())
+        self.vectors = {}
+        self.legend = widgets.Output()
+
+        import ipyvolume.pylab as p3
```
prabhuramachandran: Why the name `p3`?

pysph/tools/ipy_viewer.py
```diff
@@ -540,3 +541,294 @@ def _save_figure_handler(self, change):
                 )
                 break
         self._widgets.save_figure.value = ""
+
+
+class ParticleArrayWidgets3D(object):
+
+    def __init__(self, particlearray, particles_type):
```
prabhuramachandran: what is `particles_type`?  If you mean or the `name` of the array?

deeptavker: It's the same as we were using for 2d. It takes the values - _fluid_, _boundary_, _inlet_, etc.

pysph/tools/ipy_viewer.py
deeptavker: I saw this in a couple of examples.

pysph/tools/ipy_viewer.py
deeptavker: Yep, I wonder why pep8 did not catch that.

pysph/tools/ipy_viewer.py
prabhuramachandran: Then I would prefer that you change both the variable names to something more meaningful like `array_name` or even just `name`.  That said, you don't need to pass the attribute at all as the name is always available from the particle array itself.  `particlearray.name` will give you the same thing.  Can you please fix this?  Thanks.

prabhuramachandran:
@deeptavker -- I hope you have tested these out once as we do not have any tests for this.  Actually, it may be nice to have some tests for this eventually, perhaps using mocks.  If you can confirm that these all work, I am happy to merge for now but do keep the testing in mind for later.

deeptavker:
Tested, RTM. I'll have a look at the mock tests after Binder and documentation.

PR #32, opened 2017-09-05
Changes in geometry and test_geometry

sankasuraj:
Vectorising checks in geometry and test_geometry to reduce the time taken to generate the geometry

### DIFF
--- a/pysph/tools/tests/test_geometry.py
+++ b/pysph/tools/tests/test_geometry.py
@@ -172,57 +172,62 @@ class TestGeometry(unittest.TestCase):
         assert not np.any(zc) > (length / 2.0)
 
     def test_get_4digit_naca_airfoil(self):
-        dx = 0.05
+        dx = 0.1
         c = 1.5
         camber = '44'
-        t = 10
-        airfoil = np.random.choice(['00', camber]) + str(t)
-        x, y = G.get_4digit_naca_airfoil(dx, airfoil, c)
-        count = 0
-        t = 0.01 * t
-        for xi, yi in zip(x, y):
-            yt = 5.0 * t * (0.2969 * np.sqrt(xi / c) - 0.1260 * (xi / c) -
-                            0.3516 * ((xi / c)**2.) + 0.2843 * (xi / c)**3. -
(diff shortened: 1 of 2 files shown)


pysph/tools/tests/test_geometry.py
```diff
@@ -17,32 +11,34 @@ class TestGeometry(unittest.TestCase):
...
             n = np.random.randint(1, 10)
             mat = np.random.rand(n, n)
-            assert np.allclose(e(mat), matrix_exp(mat))
+            assert np.allclose(e(mat), G.matrix_exp(mat))
         except ImportError:
-            raise unittest.SkipTest()
+            raise unittest.SkipTest('Scipy is not installed')
 
     def test_extrude(self):
         n = np.random.randint(1, 20)
```
prabhuramachandran: Please do not use random numbers for the test unless absolutely necessary, otherwise it is very difficult to reproduce the test, otherwise set the random number seed.

pysph/tools/tests/test_geometry.py
```diff
@@ -104,98 +100,151 @@ def test_get_2d_wall(self):
...
+                else:
+                    yl = yc - yt * np.cos(theta)
+                    if yi < yl + dx:
+                        count += 1
+        assert count == 0
+
+    def test_get_5digit_naca_airfoil(self):
+        dx = 0.05
+        c = 1.5
+        series = np.random.choice(['210', '220', '230', '240', '250', '221',
```
prabhuramachandran: Do not make this random.  If you want to test them all, then test them all using a parameter or put it in a for loop.  No sense in just testing one of them each time.  The idea of a test is to cover the possible options but just enough so as to make sure everything is well tested.  You should really measure coverage of the geometry module and make sure that it is almost entirely covered.

sankasuraj:
Used random.seed and also removed random.choice in airfoils so as to test all type of airfoils available to increase coverage which is at 94% now which seems good as we are not testing the show2d and 3d functions

sankasuraj:
Removed randomness completely from tests

prabhuramachandran:
Much better, thanks!

PR #38, opened 2017-09-19
Fix brute force on GPU

adityapb:


prabhuramachandran:
Merging but I think the use_double should do the same thing that the other NNPS ones do, this can be fixed later.

PR #40, opened 2017-09-20
Fix BruteForceNNPS and ZOrderGPUNNPS for multiple particle arrays

adityapb:
I've added a test for finding neighbors multiple times.
The tests will fail because of ZOrderGPUNNPS

### DIFF
--- a/pysph/base/gpu_nnps_base.pyx
+++ b/pysph/base/gpu_nnps_base.pyx
@@ -293,6 +293,8 @@ cdef class BruteForceNNPS(GPUNNPS):
         self.dst = self.pa_wrappers[ dst_index ]
 
     cdef void find_neighbor_lengths(self, nbr_lengths):
+        # IMPORTANT NOTE: pyopencl uses the length of the first argument
+        # to determine the global work size
         arguments = \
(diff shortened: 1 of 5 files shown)


pysph/base/gpu_nnps_base.pyx
```diff
@@ -294,9 +294,9 @@ cdef class BruteForceNNPS(GPUNNPS):
 
     cdef void find_neighbor_lengths(self, nbr_lengths):
         arguments = \
-                """%(data_t)s* s_x, %(data_t)s* s_y, %(data_t)s* s_z,
-                %(data_t)s* s_h, %(data_t)s* d_x, %(data_t)s* d_y,
-                %(data_t)s* d_z, %(data_t)s* d_h, unsigned int num_particles,
+                """%(data_t)s* d_x, %(data_t)s* d_y, %(data_t)s* d_z,
```
prabhuramachandran: Good catch!  I think a small comment here and elsewhere is in order to make sure that someone reading the source remembers that the global work size is determined from the length of the first argument.

prabhuramachandran:
Excellent!  Thank you!

PR #43, opened 2017-09-23
Implemented akinci scheme. Example appended.

dineshadepu:
Implemented Akinci scheme correctly. 

I appended an example to check it. Please run the example with different density values i.e.,

change solid_rho to different values, and check it.

### DIFF
--- a/pysph/examples/rigid_body/sphere_in_vessel_akinci.py
+++ b/pysph/examples/rigid_body/sphere_in_vessel_akinci.py
@@ -4,7 +4,6 @@ Check basic equations of SPH to throw a ball inside the vessel
 """
 from __future__ import print_function
 import numpy as np
-import matplotlib.pyplot as plt
 
 # PySPH base and carray imports
 from pysph.base.utils import (get_particle_array_wcsph,
@@ -97,6 +96,7 @@ def get_density(y):
 
 
 def geometry():
+    import matplotlib.pyplot as plt
     # please run this function to know how
     # geometry looks like
     x_tank, y_tank = create_boundary()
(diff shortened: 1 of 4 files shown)


pysph/examples/rigid_body/sphere_in_vessel_akinci.py
```diff
@@ -0,0 +1,220 @@
...
+    _tmp = gamma / (rho_0 * c_0**2)
+
+    rho = np.zeros_like(y)
+    for i in range(len(rho)):
+        p_i = rho_0 * 9.81 * (height_water_clmn - y[i])
+        rho[i] = rho_0 * (1 + p_i * _tmp)**(1. / gamma)
+    return rho
+
+
+def geometry():
```
prabhuramachandran: Please use the functions provided already in the pysph.tools.geometry module instead of these.  This plotting function is also not needed here.

pysph/examples/rigid_body/sphere_in_vessel_akinci.py
```diff
@@ -0,0 +1,220 @@
+"""A sphere of density 500 falling into a hydrostatic tank (15 minutes)
+
+Check basic equations of SPH to throw a ball inside the vessel
+"""
+from __future__ import print_function
+import numpy as np
+import matplotlib.pyplot as plt
```
prabhuramachandran: Do not do this as our test suite is sometimes run on machines without matplotlib and this import may fail.  No need for this.

pysph/examples/rigid_body/sphere_in_vessel_akinci.py
```diff
@@ -0,0 +1,220 @@
...
+from pysph.sph.wc.basic import TaitEOSHGCorrection, MomentumEquation
+from pysph.solver.application import Application
+from pysph.sph.rigid_body import (
+    BodyForce, SummationDensityBoundary, CutoffDensity, RigidBodyCollision,
+    LiuFluidForce, RigidBodyMoments, RigidBodyMotion, AkinciRigidFluidCoupling,
+    RK2StepRigidBody)
+
+
+def create_boundary():
+    dx = 2
```
prabhuramachandran: These can all use pysph.tools.geometry.

pysph/examples/rigid_body/sphere_in_vessel_akinci.py
```diff
@@ -0,0 +1,220 @@
...
+        V = np.ones_like(xt) * dx * dx * 1e-6
+        tank = get_particle_array_wcsph(x=xt, y=yt, h=h, m=m, rho=rho,
+                                        rad_s=rad_s, V=V, name="tank")
+        for name in ['fx', 'fy', 'fz']:
+            tank.add_property(name)
+
+        dx = 1
+        xc, yc = create_sphere(1)
+        m = np.ones_like(xc) * self.solid_rho * dx * 1e-3 * dx * 1e-3
+        rho = np.ones_like(xc) * self.solid_rho
```
prabhuramachandran: With latest pysph you do not need ones_like for simple constants, just pass in a constant for the constant properties and it should work.  It will broadcast this to the same shape as x, y, z and use it.

pysph/sph/rigid_body.py
```diff
@@ -225,13 +225,41 @@ def initialize(self, d_idx, d_m, d_fx, d_fy, d_fz):
...
+class CutoffDensity(Equation):
+    """Limit density to real fluid density
+
+
+    """
+    def __init__(self, dest, sources, fluid_rho=1000.0):
+        self.fluid_rho = fluid_rho
+        super(CutoffDensity, self).__init__(dest, sources)
+
+    def initialize(self, d_idx, d_rho):
```
prabhuramachandran: Is this really needed?  Instead of changing the density directly, you should probably have a different equation of state that does not generate negative pressures by finding `max(rho0, d_rho[d_idx))` when computing the pressure.  Otherwise your `m_j/rho_j` terms will all be wrong.  So in short, don't change the rho but use a different EOS.

pysph/sph/rigid_body.py
```diff
@@ -326,6 +394,7 @@ def __init__(self, dest, sources):
 
     def loop(self, d_idx, d_m, d_rho, d_au, d_av, d_aw,  d_p,
              s_idx, s_V, s_fx, s_fy, s_fz, DWIJ, s_m, s_p, s_rho):
+
```
prabhuramachandran: Please don't add unnecessary whitespace.

dineshadepu:
Merge this for now If it is urgent. Or else we can close this and merge when 3d example is ready.

dineshadepu:
Closing this for now. Reopen it once 3d is done.

PR #45, opened 2017-09-27
Add DeviceArray to avoid reallocations

adityapb:


pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
     _queue = q
 
 
+class DeviceArray(cl.array.Array):
+    def __init__(self, queue, dtype, n=0):
+        self.queue = queue
```
prabhuramachandran: Don't need this, just use `get_queue` to get a suitable queue.  I don't know if we need multiple separate queues?

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,47 @@
...
+import pyopencl as cl
+
 from pysph.base.utils import get_particle_array  # noqa: E402
-from pysph.base.opencl import DeviceHelper  # noqa: E402
+from pysph.base.opencl import DeviceHelper, DeviceArray  # noqa: E402
+
+
+class TestDeviceArray(TestCase):
+    def setUp(self):
+        self.ctx = cl.create_some_context(interactive=False)
```
prabhuramachandran: Again, no need for this, if you see the other tests I don't bother with the queue, see the `TestDeviceHelper` class.

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,47 @@
...
+        self.queue = cl.CommandQueue(self.ctx)
+        self.dev_array = DeviceArray(self.queue, np.int32)
+
+    def tearDown(self):
+        self.dev_array = DeviceArray(self.queue, np.int32)
```
prabhuramachandran: Why?  What is this doing in the tearDown?

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,47 @@
...
+        assert len(self.dev_array.get_array()) == 64
+        assert self.dev_array.length == 0
+
+    def test_resize_with_reallocation(self):
+        self.dev_array.resize(64)
+        assert len(self.dev_array.get_array()) == 64
+        assert self.dev_array.length == 64
+
+    def test_resize_without_reallocation(self):
+        self.dev_array = DeviceArray(self.queue, np.int32, n=128)
```
prabhuramachandran: Use the Given/When/Then comments to make it clear what is being tested.

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,47 @@
...
+    def setUp(self):
+        self.ctx = cl.create_some_context(interactive=False)
+        self.queue = cl.CommandQueue(self.ctx)
+        self.dev_array = DeviceArray(self.queue, np.int32)
+
+    def tearDown(self):
+        self.dev_array = DeviceArray(self.queue, np.int32)
+
+    def test_reserve(self):
+        self.dev_array.reserve(64)
```
prabhuramachandran: I don't think the setUp method is doing much, instead just create the `DeviceArray` here and use that.

pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
...
+        self.length = n
+        if n == 0:
+            n = 16
+        self.alloc = n
+        self.dtype = dtype
+        self._array = cl.array.empty(queue, n, dtype)
```
prabhuramachandran: Why is this alone with an `_array`?

prabhuramachandran: Instead maybe this should be `_data` and the property should really be `array` because that is what it is, right?  This `_array` is just the block of data allocated but the array is a suitable view.

pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
...
+
+    @property
+    def data(self):
+        return self._array[:self.length]
```
prabhuramachandran: I would suggest that the view only be created when the length changes, otherwise you are still creating a new object every time someone does `dev_array.data`.

pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
...
+
+    def resize(self, size):
+        self.reserve(size)
+        self.length = size
+
+    def reserve(self, size):
+        if size > self.alloc:
+            self._array = cl.array.empty(self.queue, size, self.dtype)
+            self.alloc = size
```
prabhuramachandran: The semantics of reserve with STL and other libraries is that it does not destroy existing data.  So you should try to retain whatever old data there already is, in your case you are blowing it away.  See the device helper code...

pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
...
+
+    def get_array(self):
+        return self._array
+
+    def set_array(self, array):
```
prabhuramachandran: I don't really like the name too much as it isn't really clear what it is doing.  It isn't evident that it is really replacing the underlying data!  Usually set suggests that you are simply setting the values.  Is this really needed?  If you wanted to set a new array, you might as well have array or data as a kwarg to `__init__` making it  obvious and eliminate the need for this method

pysph/base/opencl.py
```diff
@@ -35,6 +35,47 @@ def set_queue(q):
...
+    def resize(self, size):
+        self.reserve(size)
+        self.length = size
+
+    def reserve(self, size):
+        if size > self.alloc:
+            self._array = cl.array.empty(self.queue, size, self.dtype)
+            self.alloc = size
+
+    def get_array(self):
```
prabhuramachandran: Where is this used?  Why is it needed?

pysph/base/opencl.py
prabhuramachandran: On second thought, I can see the use of this when someone has a reference to this object, so this can stay.

pysph/base/opencl.py
adityapb: This is used only for testing

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,55 @@
...
 from pysph.base.utils import get_particle_array  # noqa: E402
-from pysph.base.opencl import DeviceHelper  # noqa: E402
+from pysph.base.opencl import DeviceHelper, DeviceArray, \
+        get_queue  # noqa: E402
+
+
+class TestDeviceArray(TestCase):
+    def test_reserve(self):
+        # Given
+        self.dev_array = DeviceArray(np.int32, n=16)
```
prabhuramachandran: Well, you don't even need `self.dev_array` as you can just do `dev_array = ...` and the code becomes even simpler!

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,55 @@
...
+        # When
+        self.dev_array.reserve(64)
+        # Then
+        assert len(self.dev_array.get_data()) == 64
+        assert self.dev_array.length == 16
+        assert self.dev_array.array[0] == 1
+
+    def test_resize_with_reallocation(self):
+        # Given
+        self.dev_array = DeviceArray(np.int32)
```
prabhuramachandran: Same here and in all the code below.

pysph/base/tests/test_opencl.py
```diff
@@ -4,8 +4,55 @@
...
+        # When
+        self.dev_array.resize(64)
+        # Then
+        assert len(self.dev_array.get_data()) == 64
+        assert self.dev_array.length == 64
```
prabhuramachandran: You should also test that the content of the old data is preserved after resize.

pysph/base/opencl.py
```diff
@@ -35,6 +35,52 @@ def set_queue(q):
     _queue = q
 
 
+class DeviceArray(cl.array.Array):
+    def __init__(self, dtype, n=0):
+        self.queue = get_queue()
+        self.length = n
+        if n == 0:
+            n = 16
+        self.alloc = n
```
prabhuramachandran: Hmm, can you reuse `self.set_data` here to do this and remove some duplicated code?

PR #47, opened 2017-10-03
Add GPUDomainManager and profiling option for OpenCL

adityapb:


adityapb:
Maybe we should add class variables of CPUDomainManager as properties in DomainManager?

pysph/base/opencl.py
```diff
@@ -1,14 +1,18 @@
...
 import pyopencl as cl
 import pyopencl.array  # noqa: 401
+from collections import defaultdict
+from operator import itemgetter
 
 from .config import get_config
 
 _ctx = None
 _queue = None
+_profile_info = defaultdict()
```
prabhuramachandran: Just use `defaultdict(float)`

pysph/base/opencl.py
```diff
@@ -35,6 +42,31 @@ def set_queue(q):
     _queue = q
 
 
+def profile(name, event):
+    global _profile_info
+    event.wait()
+    time = (event.profile.end - event.profile.start) * 1e-9
+    if name in _profile_info:
+        _profile_info[name] += time
+    else:
```
prabhuramachandran: No need for the if block, since the defaultdict will always give you a default 0 float, you can just have `_profile_info[name] += time`

pysph/base/opencl.py
```diff
@@ -35,6 +42,31 @@ def set_queue(q):
...
+        _profile_info[name] = time
+
+
+def profile_info():
```
prabhuramachandran: Don't like this, name, merge the `profile_info` and `print_profile` into one function saying `print_profile()`.

pysph/base/opencl.py
```diff
@@ -35,6 +42,31 @@ def set_queue(q):
...
+    global _profile_info
+    _profile_info = sorted(_profile_info.iteritems(), key=itemgetter(1),
+                           reverse=True)
+    print_profile(_profile_info)
+
+
+def print_profile(info):
+    if len(info) == 0:
+        return
```
prabhuramachandran: Instead print "no profile information available." or so, as you are promising to print something...

pysph/base/z_order_gpu_nnps.pyx
```diff
@@ -79,16 +81,22 @@ cdef class ZOrderGPUNNPS(GPUNNPS):
...
 
         pa_gpu = pa_wrapper.pa.gpu
-        fill_pids(pa_gpu.x, pa_gpu.y, pa_gpu.z,
-                self.cell_size, self.xmin[0], self.xmin[1], self.xmin[2],
+        event = fill_pids(pa_gpu.x, pa_gpu.y, pa_gpu.z,
+                self.cell_size,
+                self.make_vec(self.xmin[0], self.xmin[1], self.xmin[2]),
                 self.pid_keys[pa_index].array, self.pids[pa_index].array)
 
+        if self._profile:
```
prabhuramachandran: Instead, should we just use a decorator so we say 
```
fill_pids = profile(self.helper.get_kernel("fill_pids"))
```
and just go about calling `fill_pids`?

pysph/base/z_order_gpu_nnps.pyx
```diff
@@ -119,22 +132,25 @@ cdef class ZOrderGPUNNPS(GPUNNPS):
...
             self.cids[pa_index].array, self.cid_to_idx[pa_index].array
         )
 
+        if self._profile:
+            profile("map_cid_to_idx", event)
+
         fill_cids = self.helper.get_kernel("fill_cids")
 
-        fill_cids(self.pid_keys[pa_index].array, self.cids[pa_index].array,
+        event = fill_cids(self.pid_keys[pa_index].array, self.cids[pa_index].array,
```
prabhuramachandran: A lot of this will be much nicer with a profile -- in fact, you can do this in `helper.get_kernel` directly.

pysph/sph/acceleration_eval_opencl_helper.py
```diff
@@ -81,9 +82,15 @@ def _call_kernel(self, info, extra_args):
                 cache._start_idx_gpu.array.data,
                 cache._neighbors_gpu.array.data
             ] + extra_args
-            call(*args)
+            event = call(*args)
+            if self._profile:
+                profile(call.get_info(cl.kernel_info.FUNCTION_NAME),
```
prabhuramachandran: I think `call.function_name` will also work just fine instead of this additional `call.get_info` method.

pysph/sph/acceleration_eval_opencl_helper.py
```diff
@@ -81,9 +82,15 @@ def _call_kernel(self, info, extra_args):
...
+                        event)
         else:
-            call(*(args + extra_args))
+            event = call(*(args + extra_args))
+            if self._profile:
```
prabhuramachandran: Again, if profile was a decorator or so we could either do the profiling or just return the original if not profiling.

pysph/base/opencl.py
```diff
@@ -35,6 +42,36 @@ def set_queue(q):
...
+    print("{:<30} {:<30}".format('Kernel', 'Time'))
+    tot_time = 0
+    for kernel, time in _profile_info:
+        print("{:<30} {:<30}".format(kernel, time))
+        tot_time += time
+    print("Total profiled time: %g secs" % tot_time)
+
+
+def profile_kernel(kernel, name):
+    def _profile_knl(*args):
```
prabhuramachandran: You can put the if outside so if profiling is on, it just returns the original function and if not, it calls profile after the kernel invocation.  Also, you could return the event from the wrapped function just in case someone wants to use it to wait or something.

pysph/base/stratified_sfc_gpu_nnps.pyx
```diff
@@ -102,15 +105,16 @@ cdef class StratifiedSFCGPUNNPS(GPUNNPS):
     cpdef _bin(self, int pa_index):
         cdef NNPSParticleArrayWrapper pa_wrapper = self.pa_wrappers[pa_index]
 
-        fill_pids = self.helper.get_kernel("fill_pids")
+        fill_pids = profile_kernel(*self.helper.get_kernel("fill_pids"))
```
prabhuramachandran: Do you want the helper to always return a profiled kernel?

prabhuramachandran:
Made a few more comments, but I can do this too later.

pysph/base/opencl.py
```diff
@@ -35,6 +42,38 @@ def set_queue(q):
...
+        event = kernel(*args)
+        profile(name, event)
+        return event
+    if get_config().profile:
+        return _profile_knl
+    return kernel
```
prabhuramachandran: Put this return in an else to make it easier to read and reason.

pysph/base/stratified_sfc_gpu_nnps.pyx
prabhuramachandran: These should not call `profile_kernel` as the helper only returns a profiled kernel, right?

pysph/base/stratified_sfc_gpu_nnps.pyx
```diff
@@ -163,16 +170,16 @@ cdef class StratifiedSFCGPUNNPS(GPUNNPS):
...
                 self.src.get_number_of_particles(), self.pid_keys[self.src_index],
                 self.pids[self.dst_index], self.pids[self.src_index], nbr_lengths,
                 self.radius_scale, self.hmin, self.interval_size,
                 self.start_idx_levels[self.src_index],
                 self.max_num_bits, self.num_levels, self.num_particles_levels[self.src_index])
 
     cdef void find_nearest_neighbors_gpu(self, nbrs, start_indices):
-        find_nbrs = self.helper.get_kernel("find_nbrs",
-                sorted=self._sorted)
+        find_nbrs = profile_kernel(*self.helper.get_kernel("find_nbrs",
```
prabhuramachandran: This should just be `self.helper.get_kernel(...)`

PR #50, opened 2017-10-05
Rigidfluid bug

dineshadepu:
3d example implemented.


dineshadepu:
Implemented 3d example of three spheres falling into a vessel. 

Introduction of examples changed.

Check if it is ok.

pysph/sph/rigid_body.py
```diff
@@ -242,6 +242,20 @@ def initialize(self, d_idx, d_m, d_fx, d_fy, d_fz, d_num_body, d_mi):
         d_fz[d_idx] = d_m[d_idx]*self.gz
 
 
+class SummationDensityBoundary(Equation):
+    r"""Good old Summation density:
```
prabhuramachandran: Docstring is not correct.

PR #51, opened 2017-10-11
Use DeviceArray in DeviceHelper

adityapb:


pysph/base/opencl.py
```diff
@@ -163,11 +160,10 @@ def add_prop(self, name, carray):
         particle array.
         """
         np_array = self._get_array(carray)
-        g_ary = cl.array.empty(self._queue, carray.alloc, np_array.dtype)
-        view = g_ary[:carray.length]
-        view.set(np_array)
+        g_ary = DeviceArray(np_array.dtype, n=carray.length)
+        g_ary.array.set(np_array)
```
prabhuramachandran: I think DeviceArray should not be a subclass of `cl.array.Array`, as in that case we should be able to just do `g_ary.set(np_array)`, I doubt that this is what you intend.

prabhuramachandran:
LGTM, thanks.

PR #58, opened 2017-10-23
OpenMP Schedule

rahulgovind:
Allow users to specify which OpenMP schedule to use through the `--omp-schedule` parameter. The default value is set to `dynamic` with a chunksize of `64`.

### DIFF
--- a/pysph/base/config.py
+++ b/pysph/base/config.py
@@ -10,6 +10,7 @@ class Config(object):
         self._use_openmp = None
         self._use_opencl = None
         self._use_double = None
+        self._omp_schedule = None
         self._profile = None
 
     @property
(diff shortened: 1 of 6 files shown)


pysph/base/config.py
```diff
@@ -26,6 +27,34 @@ def _use_openmp_default(self):
...
+    def omp_schedule(self):
+        if self._omp_schedule is None:
+            self._omp_schedule = self._omp_schedule_default()
+        return self._omp_schedule
+
+    @omp_schedule.setter
+    def omp_schedule(self, value):
+        if len(value) != 2 or \
+           value[0].lower() not in ("static", "dynamic", "guided"):
+            raise ValueError("Invalid OpenMP Schedule")
```
prabhuramachandran: Please also specify the supplied schedule in the error message.

PR #62, opened 2017-12-03
2d viewer fix - modified the code to work with ipympl backend

deeptavker:
- ipympl backend is now mandatory

### DIFF
--- a/pysph/tools/ipy_viewer.py
+++ b/pysph/tools/ipy_viewer.py
@@ -1,10 +1,11 @@
 import json
 import glob
 from pysph.solver.utils import load, get_files
-from IPython.display import display, clear_output, Image
+from IPython.display import display, Image
 import ipywidgets as widgets
 import matplotlib.pyplot as plt
 import matplotlib as mpl
+import numpy as np
 
 
 class Viewer(object):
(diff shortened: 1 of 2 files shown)


pysph/tools/ipy_viewer.py
```diff
@@ -356,150 +357,206 @@ def _frame_handler(self, change):
...
+                        v1,
+                        v2,
+                        vmag,
+                        scale=pa_widgets.vector_scale.value,
+                        width=(pa_widgets.vector_width.value)/10000,
+                    )
+            self._vector_ax.set_xlim(self._scatter_ax.get_xlim())
+            self._vector_ax.set_ylim(self._scatter_ax.get_ylim())
+
+        except AttributeError:
```
prabhuramachandran: This try except seems rather strange to me and I don't like it.  It is not at all obvious why there should be an attribute error and why it is OK to just pass on it? Can you please elaborate?

deeptavker: So, the vector widget takes input "var1,var2" where var1 and var2 should each be an output property of the respective particlearray (say "u,v" or "au,av"), in any other case (say the input is "foo,foo1"), an AttributeError is encountered and to circumvent that, I added the try-except block. `pass` because on entering any invalid input, it doesn't do anything. If need be, I can add a TextArea widget at the bottom to display all such errors in a user-friendly language.

prabhuramachandran: My issue with this is that the try block should enclose the smallest possible code where you anticipate the error and not a blanket one where someone is wondering why you need to catch an exception for all of that code.  Please modify it along these lines.

PR #64, opened 2018-01-10
Small fixes in VTK export

nilsmeyerkit:
Hi,

I noticed that 'dump_vtk' fails for some cases when using TVF. I suggest the following fixes for this issue.

Nils


prabhuramachandran:
Thanks, this looks good although it would have been nicer with tests, I'll merge it now though.  Thanks!

PR #66, opened 2018-01-24
STL Converter

rahulgovind:


rahulgovind:
I have added `numpy-stl`, the package used by the converter to parse STL files, to`requirements.txt` since it's required for the tests. It seems like I'm missing something. Is there a way to avoid this and only require the package to be installed during the tets?

prabhuramachandran:
Well, I don't have a problem requiring numpy-stl per se but yes I can see that it may be desirable to have a separate set of packages required for testing.  setuptools has a `tests-require` keyword argument that can be used but that only works if you invoke the tests as `python setup.py test`.  Maybe we could have a `requirements-test.txt` and add that to the tox.ini and appveyor.yml?

pysph/tools/geometry_stl.pyx
```diff
@@ -0,0 +1,190 @@
...
+
+
+def fill_triangle(triangle, h=0.1):
+    EPS = np.finfo(float).eps
+    if norm(triangle[0] - triangle[1]) < EPS or \
+       norm(triangle[1] - triangle[2]) < EPS or \
+       norm(triangle[2] - triangle[0]) < EPS:
+        raise ValueError("Zero area triangle")
+
+    """
```
prabhuramachandran: I prefer these to be comments and not free-floating strings.

pysph/tools/geometry_stl.pyx
```diff
@@ -0,0 +1,190 @@
...
+    return np.dot(x - p, a), np.dot(x - p, b)
+
+
+def fill_triangle(triangle, h=0.1):
+    EPS = np.finfo(float).eps
+    if norm(triangle[0] - triangle[1]) < EPS or \
+       norm(triangle[1] - triangle[2]) < EPS or \
+       norm(triangle[2] - triangle[0]) < EPS:
+        raise ValueError("Zero area triangle")
```
prabhuramachandran: Perhaps make this error a bit more useful by passing `str(triangle)` as well?

pysph/tools/geometry_stl.pyx
```diff
@@ -0,0 +1,190 @@
...
+    m = mesh.Mesh.from_file(stl_fname)
+    x_list, y_list, z_list = [], [], []
+    for i in range(len(m.vectors)):
+        try:
+            x1, y1, z1 = fill_triangle(m.vectors[i], dx_triangle)
+            x2, y2, z2 = get_triangle_sides(m.vectors[i], dx_triangle)
+            x_list.append(np.r_[x1, x2])
+            y_list.append(np.r_[y1, y2])
+            z_list.append(np.r_[z1, z2])
+        except Exception as e:
```
prabhuramachandran: Don't catch all exceptions, just catch ValueError, in fact, create a ZeroTriangleError exception of your own and catch that.

pysph/tools/geometry_stl.pyx
```diff
@@ -0,0 +1,190 @@
...
+            continue
+
+    x = np.concatenate(x_list)
+    y = np.concatenate(y_list)
+    z = np.concatenate(z_list)
+
+    return x, y, z
+
+
+def get_stl_surface(stl_fname, dx_sph, h_sph, radius_scale=1.0,
```
prabhuramachandran: Since this seems like the only public function you want users to use, maybe the rest could all start with an `_`?

pysph/tools/geometry_stl.pyx
prabhuramachandran: This stl mesh will have many coincident points, is that correct?

rahulgovind: Indeed.

prabhuramachandran:
LGTM, thanks.

PR #67, opened 2018-01-28
Separate test requirements

rahulgovind:
One of the possible ways to separate test dependencies from the core installation dependencies. Here, the test dependencies are added to `extra_requires` under the `test` key. This way, the test dependencies can be installed by running `pip install .[test]`. 

Using this approach, requirements.txt will no longer be required but I haven't removed it in this commit since it would break any other build/installation scripts running `pip install -r requirements.txt`.

Edit: requirements.txt would also be useful if the dependencies have to be installed without installing the package.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ install:
   - edm install -y numpy cython pytest mock h5py
 
   # Install pysph related dependencies.
-  - edm run -- pip install -r requirements.txt
+  - edm run -- pip install .[test]
 
   # Build pysph.
   - edm run -- python setup.py develop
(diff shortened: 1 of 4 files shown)


prabhuramachandran:
Please merge and fix the conflicts.

appveyor.yml
```diff
@@ -22,7 +22,7 @@ install:
   - edm install -y numpy cython pytest mock h5py
 
   # Install pysph related dependencies.
-  - edm run -- pip install -r requirements.txt
+  - edm run -- pip install .[test]
```
prabhuramachandran: Does this also install the package (i.e. pysph) or does it just install the deps?

rahulgovind: Right. It installs the package too. The next statement is redundant.

PR #68, opened 2018-01-29
Added missing 'else' statement in VTK export.

nilsmeyerkit:
I just realized that I did not test VTK export with given sclarars using the option "-s ...". This, obviously, needs the missing 'else' statement.

Sorry for the inconvinience and not providing tests.


prabhuramachandran:
Thanks!

PR #71, opened 2018-02-20
MLS first order functions

sankasuraj:


pysph/solver/tools.py
```diff
@@ -123,8 +123,8 @@ def __init__(self, app, arr_names, corr='shepard', freq=10, kernel=None):
         self.particles = app.particles
         self.arrs = [get_array_by_name(self.particles, i) for i in self.names]
         try:
-            assert self.corr in ['shepard']
-        except:
+            assert self.corr in ['shepard', 'mls2d_1', 'mls3d_1']
```
prabhuramachandran: You can simply do:
```
options = ['shepard', ...]
assert self.corr in ..., 'Corr should be one of %s' % options
```

pysph/sph/tests/test_gjsolve.py
```diff
@@ -0,0 +1,23 @@
+from pysph.sph.wc.density_correction import gj_Solve
+import numpy as np
+import unittest
+
+
+class TestGjSolve(unittest.TestCase):
+
+    def test_gj_solve(self):
+        for i in range(10):
```
prabhuramachandran: Random matrices make it hard to test and debug as the matrices change each time.  Please use a more deterministic test.  Use specific matrices, also pick bad matrices and see what it does.

pysph/solver/tools.py
```diff
@@ -151,9 +151,65 @@ def _get_sph_eval_shepard(self):
...
+                MLSFirstOrderPreStep2D, MLSFirstOrder)
+            from pysph.tools.sph_evaluator import SPHEvaluator
+            from pysph.sph.equation import Group
+            arrs = self.arrs
+            eqns = []
+            n = 3
+            for arr in arrs:
+                lenga = len(arr.x) * n * n
+                lengb = len(arr.x) * n
+                arr.add_constant('a', [0.] * lenga)
```
prabhuramachandran: This name is a bit too generic: 'a', 'b' , please use something a bit more meaningful.  Also are these really needed?  Do you really need to store the matrix for each particle?

prabhuramachandran: Maybe we can revisit this when we add support to handle all neighbors.

pysph/sph/wc/density_correction.py
```diff
@@ -32,3 +111,140 @@ def initialize(self, d_idx, d_rho):
...
+            d_a[n * n * d_idx + i] = 0.0
+
+    def loop(self, s_idx, d_idx, d_a, s_m, s_rho, WIJ, XIJ):
+        n = declare('int')
+        n = 3
+        i = declare('int')
+        j = declare('int')
+        for i in range(n):
+            for j in range(n):
+                fac1 = 1.0 if i == 0 else XIJ[i - 1]
```
prabhuramachandran: Can you comment on this or write out the equations used here?

pysph/sph/wc/density_correction.py
```diff
@@ -32,3 +111,140 @@ def initialize(self, d_idx, d_rho):
...
+            d_a[n * n * d_idx + i] = 0.0
+
+    def loop(self, s_idx, d_idx, d_a, s_m, s_rho, WIJ, XIJ):
+        n = declare('int')
+        n = 4
+        i = declare('int')
+        j = declare('int')
+        for i in range(n):
+            for j in range(n):
+                fac1 = 1.0 if i == 0 else XIJ[i - 1]
```
prabhuramachandran: I don't think the ternary operator will work on the GPU.  This should be easy to support but isn't currently.

prabhuramachandran: Also move the conditional outside the for j loop as it doesn't need to be evaluated 9 times and only 3 since it is independent of j.

prabhuramachandran: Do this for all similar cases in your code please.

sankasuraj:
Modified Density reinitialization equations using loop_all function

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,285 @@
...
+        x = d_x[d_idx]
+        y = d_y[d_idx]
+        z = d_z[d_idx]
+        for i in range(N_NBRS):
+            s_idx = NBRS[i]
+            xij[0] = x - s_x[s_idx]
+            xij[1] = y - s_y[s_idx]
+            xij[2] = z - s_z[s_idx]
+            rij = sqrt(xij[0] * xij[0] + xij[1] * xij[1] + xij[2] * xij[2])
+            wij = KERNEL.kernel(xij, rij, s_h[s_idx])
```
prabhuramachandran: shouldn't this be `(d_h[d_idx] + s_h[s_idx])*0.5`

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,285 @@
...
+        b1 = res[1]
+        b2 = res[2]
         d_rho[d_idx] = 0.0
+        for k in range(N_NBRS):
+            s_idx = NBRS[k]
+            xij[0] = x - s_x[s_idx]
+            xij[1] = y - s_y[s_idx]
+            xij[2] = 0.
+            rij = sqrt(xij[0] * xij[0] + xij[1] * xij[1])
+            wij = KERNEL.kernel(xij, rij, s_h[s_idx])
```
prabhuramachandran: shouldn't this be `(d_h[d_idx] + s_h[s_idx])*0.5`

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,285 @@
...
+        for i in range(n):
+            for j in range(n):
+                amls[n * i + j] = 0.0
+        for k in range(N_NBRS):
+            s_idx = NBRS[k]
+            xij[0] = x - s_x[s_idx]
+            xij[1] = y - s_y[s_idx]
+            xij[2] = z - s_z[s_idx]
+            rij = sqrt(xij[0] * xij[0] + xij[1] * xij[1] + xij[2] * xij[2])
+            wij = KERNEL.kernel(xij, rij, s_h[s_idx])
```
prabhuramachandran: shouldn't this be `(d_h[d_idx] + s_h[s_idx])*0.5`

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,285 @@
...
+        b2 = res[2]
+        b3 = res[3]
+        d_rho[d_idx] = 0.0
+        for k in range(N_NBRS):
+            s_idx = NBRS[k]
+            xij[0] = x - s_x[s_idx]
+            xij[1] = y - s_y[s_idx]
+            xij[2] = z - s_z[s_idx]
+            rij = sqrt(xij[0] * xij[0] + xij[1] * xij[1] + xij[2] * xij[2])
+            wij = KERNEL.kernel(xij, rij, s_h[s_idx])
```
prabhuramachandran: shouldn't this be `(d_h[d_idx] + s_h[s_idx])*0.5`

pysph/sph/tests/test_gjsolve.py
```diff
@@ -0,0 +1,79 @@
...
+               [0., 0., 100., 200.]]
+        b = [0.02, 1., 4., 800.]
+        result = [0.0] * n
+        gj_Solve(np.ravel(mat), b, n, result)
+        mat = np.matrix(mat)
+        new_b = mat * np.transpose(np.matrix(result))
+        new_b = np.ravel(np.array(new_b))
+        assert np.allclose(new_b, np.array(b))
+
+    def testBandMatrix(self):
```
prabhuramachandran: Again can you change the name.

prabhuramachandran:
One minor comment, I have made it possible to make multiple declarations in one line using
`x, y = declare('int')`, you could simplify some of the code with that I guess.

sankasuraj:
Doing i, j = declare('int') will work in the equation but in the gj_solve function it won't work while testing as the declare function while testing is imported from pysph.sph.equation.

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,285 @@
...
+    def initialize(self, d_idx, d_rho, d_rhotmp):
+        d_rhotmp[d_idx] = d_rho[d_idx]
 
-    def loop(self, d_idx, d_tmp_w, s_m, s_rho, s_idx, WIJ):
-        d_tmp_w[d_idx] += WIJ * s_m[s_idx] / s_rho[s_idx]
+    def loop_all(self, d_idx, d_rho, d_x, d_y, d_z, s_m, s_rhotmp, s_x, s_y,
+                 s_z, s_h, KERNEL, NBRS, N_NBRS):
+        i = declare('int')
+        s_idx = declare('long')
+        xij = declare('matrix((3,))')
```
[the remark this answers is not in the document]
sankasuraj: That doesn't seem to work for some reason

prabhuramachandran:
Ahh, yes, you will need to change the declare function, I forgot all about that. I can do that later today or you could try too.

prabhuramachandran:
We could change the syntax to `declare("type", number)` with number defaulting to 1 but if not one it returns that many copies.   There may be other hacky ways to do this but I am not in favor of doing something hacky in the Python implementation. Thoughts?

pysph/sph/wc/density_correction.py
```diff
@@ -1,34 +1,270 @@
...
+    r""" A gauss-jordan method to solve an augmented matrix for the
+        unknown variables, x, in Ax = b.
+
+    References
+    ----------
+    https://ricardianambivalence.com/2012/10/20/pure-python-gauss-jordan
+    -solve-ax-b-invert-a/
+    """
+
+    i = declare('int')
```
prabhuramachandran: These will all work in one line for the int case.

prabhuramachandran:
@sankasuraj  -- Thanks for the changes.  Is this ready to be merged or do you have anything else to add?

sankasuraj:
Ready to merge

prabhuramachandran:
I forgot, can you please add these new density filters to all of the relevant examples via a suitable command line option?  ``--density-filter `` and use the choices option https://docs.python.org/3/library/argparse.html#choices

sankasuraj:
I didn't understand properly how to add to the relevant examples, I will ask you about it in the next meeting.

PR #83, opened 2018-03-05
Separate test requirements

rahulgovind:
Listing packages required for testing separately in `requirements-test.txt`


requirements.txt
```diff
@@ -2,6 +2,4 @@ numpy
 setuptools>=6.0
 Cython>=0.20,<0.27
```
prabhuramachandran: Can you remove this requirement that I had added temporarily a while back?  So Cython >=0.20 should work.

prabhuramachandran: @rahulgovind -- can you address this please.

PR #86, opened 2018-03-08
Change in Smoothed Color Equation in Surface Tension

vikaskurapatibat:
Removed post loop of Smoothed Color Equation. If users don't want to smooth color in the problem, don't use Smoothed Color Equation.

### DIFF
--- a/pysph/sph/surface_tension.py
+++ b/pysph/sph/surface_tension.py
@@ -30,7 +30,7 @@ class SmoothedColor(Equation):
     particle.
 
     """
-    def __init__(self, dest, sources, smooth=False):
+    def __init__(self, dest, sources):
         self.smooth = smooth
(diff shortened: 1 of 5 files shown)


pysph/sph/surface_tension.py
```diff
@@ -18,34 +18,30 @@
...
+
         c_a = \sum_b \frac{m_b}{\rho_b} c_b^i \nabla_a W_{ab}\,,
 
     where, :math:`c_b^i` is the color index associated with a
     particle.
 
     """
-    def __init__(self, dest, sources, smooth=False):
-        self.smooth = smooth
+    def __init__(self, dest, sources):
```
prabhuramachandran: The `__init__` method can simply be removed in this case.

PR #88, opened 2018-03-09
Allow repeated helper functions.

prabhuramachandran:
Given the increasing number of helper functions used in equations, we
now generate the helper code better to prevent repeated definitions
resulting in compilation errors.  Equations can therefore be safely
repeated and can reuse existing helpers.


prabhuramachandran:
@sankasuraj -- this should be useful for you.  Will merge when the tests pass.

PR #91, opened 2018-04-11
Kernel Correction

sankasuraj:


pysph/examples/kernelcorrection_simulation.py
```diff
@@ -0,0 +1,90 @@
...
+    h_fluid = 2.0
+    gamma = 7.0
+    alp = 0.2
+    ro = 100.0
+    co = 10.0 * np.sqrt(2.0 * 9.81 * h_fluid)
+    dx = 0.05
+    hd = 1.5
+    h0 = dx * hd
+    dt = 0.3 * h0 / co
+    f, d = get_dam_geometry(dx, dx, hdx=hd, h_f=h_fluid, r_fluid=ro, r_tank=ro)
```
prabhuramachandran: Move this to the create_particles method.

pysph/sph/wc/kernel_correction.py
```diff
@@ -0,0 +1,175 @@
+from pysph.sph.equation import Equation
+from pysph.sph.wc.density_correction import gj_solve
+
```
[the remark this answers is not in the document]
sankasuraj: No need to import declare actually, so removed it

sankasuraj:
Using 0.1EPS instead of tol and also changed if N_NBRS>7 with a predefined factor for every problem

sankasuraj:
Ready for merging

pysph/sph/wc/crksph.py
```diff
@@ -0,0 +1,164 @@
...
+        x = d_x[d_idx]
+        y = d_y[d_idx]
+        z = d_z[d_idx]
+        h = d_h[d_idx]
+        i, j, k, s_idx, d = declare('int', 5)
+        alp, bet, gam, phi, psi = declare('int', 5)
+        xij = declare('matrix(3)')
+        dwij = declare('matrix(3)')
+        d = self.dim
+        tol = 1.0e-10
```
prabhuramachandran: Why this arbitrary number 1e-10?

pysph/examples/dam_break_2d.py
```diff
@@ -26,31 +32,40 @@
...
         self.hdx = 1.3
         group.add_argument(
             "--h-factor", action="store", type=float, dest="h_factor",
             default=1.0,
             help="Divide default h by this factor to change resolution"
         )
+        group.add_argument(
+            "--kernel-corr", action="store", type=str, dest='kernel_corr',
+            default='', help="Type of Kernel Correction"
+        )
```
prabhuramachandran: You can give a set of choices for this argument.  See this for example: https://github.com/pypr/pysph/blob/master/pysph/solver/application.py#L458

pysph/examples/dam_break_2d.py
```diff
@@ -146,7 +195,23 @@ def create_particles(self):
         if self.options.scheme == 'iisph':
             # the default position tends to cause the particles to be pushed
             # away from the wall, so displacing it by a tiny amount helps.
-            fluid.x += self.dx/4
+            fluid.x += self.dx / 4
+
+        # Adding extra properties for kernel correction
```
prabhuramachandran: Can these be added only if the correction is required instead of always?

pysph/sph/wc/kernel_correction.py
```diff
@@ -0,0 +1,198 @@
...
+            xij[0] = x - s_x[s_idx]
+            xij[1] = y - s_y[s_idx]
+            xij[2] = z - s_z[s_idx]
+            hij = (h + s_h[s_idx]) * 0.5
+            r = sqrt(xij[0] * xij[0] + xij[1] * xij[1] + xij[2] * xij[2])
+            KERNEL.gradient(xij, r, hij, dwij)
+            dw = sqrt(dwij[0] * dwij[0] + dwij[1] * dwij[1]
+                      + dwij[2] * dwij[2])
+            V = s_m[s_idx] / s_rho[s_idx]
+            if r >= 1.0e-09:
```
prabhuramachandran: How did you choose 1e-9?

pysph/sph/wc/kernel_correction.py
```diff
@@ -0,0 +1,198 @@
...
+            xij[2] = z - s_z[s_idx]
+            hij = (h + s_h[s_idx]) * 0.5
+            r = sqrt(xij[0] * xij[0] + xij[1] * xij[1] + xij[2] * xij[2])
+            KERNEL.gradient(xij, r, hij, dwij)
+            for i in range(n):
+                dwij1[i] = (dwij[i] - numerator[i] / den) / den
+            dw = sqrt(dwij1[0] * dwij1[0] + dwij1[1]
+                      * dwij1[1] + dwij1[2] * dwij1[2])
+            V = s_m[s_idx] / s_rho[s_idx]
+            if r >= 1.0e-09:
```
prabhuramachandran: Same here?

pysph/examples/dam_break_2d.py
sankasuraj: Added if conditions for each type of kernel correction

pysph/examples/dam_break_2d.py
sankasuraj: Modified this

pysph/sph/wc/crksph.py
sankasuraj: This is to prevent the denominator to be 0, changed this to 1.0e-12 to make it further small

pysph/sph/wc/kernel_correction.py
sankasuraj: This is to prevent r being 0, changed this to 1.0e-12.

pysph/sph/wc/kernel_correction.py
sankasuraj: Changed this too

sankasuraj:
Made the changes according to the comments and added doc string to crksph

prabhuramachandran:
@sankasuraj -- I just changed the code to use SPH_KERNEL as I have changed the API for that.  I will merge this once the tests pass.

prabhuramachandran:
@sankasuraj -- my major concern is that this code is not tested at all.  It will be nice to have a simple test case to ensure that all of these are functioning as desired.  Otherwise making any changes will have no effect.  For now I will merge this but this is worth thinking about.

PR #97, opened 2018-04-22
Pcisph

dineshadepu:
Dam break works with current implementation. Any additional cases will add in future.


dineshadepu:
By running pysph test gives stl module error.

prabhuramachandran:
@dineshadepu -- You need to add an `__init__.py` in `examples/pcisph` which is why the test is failing on appveyor.

PR #107, opened 2018-06-28
Shallow water equations

dileepmenon:
Added equations for solving shallow water problems. It consists of variable-h formulation, particle splitting/merging, bed surface discretization, SPH stabilization, bed friction, open boundary inflow/outflow, closed boundary treatment, etc.  Also added a particle array specific to the case of shallow water flows.


pysph/sph/swe/basic.py
```diff
@@ -0,0 +1,1677 @@
...
+        self.x_min = x_min
+        self.x_max = x_max
+        self.y_min = y_min
+        self.y_max = y_max
+        super(CheckForParticlesToSplit, self).__init__(dest, None)
+
+    def initialize(self, d_pa_to_split, d_idx):
+        d_pa_to_split[d_idx] = 0.0
+
+    def post_loop(self, d_idx, d_rho0, d_A, d_h, d_x, d_y, d_pa_to_split):
```
prabhuramachandran: You can do this in one step right in the initialize by setting the value to 1 if the condition satisfies and to 0 if not.  You then avoid an unnecessary loop!

pysph/sph/swe/basic.py
```diff
@@ -0,0 +1,1677 @@
...
+        if (d_A[d_idx] > self.A_max and d_h[d_idx] < self.h_max
+           and (self.x_min < d_x[d_idx] < self.x_max)
+           and (self.y_min < d_y[d_idx] < self.y_max)):
+            d_pa_to_split[d_idx] = 1
+
+
+class ParticleSplit(object):
```
prabhuramachandran: Where is this class used?  Why is this here?  If this is an example or something to test with, I am not sure it belongs here.  This file should generally be for equations.

pysph/sph/swe/basic.py
```diff
@@ -0,0 +1,1677 @@
...
+        """
+        self.pa_arr = pa_arr
+
+        # Ratio of mass of daughter particle located at center of hexagon to
+        # that of its parents mass
+        self.center_pa_mass_frac = 0.178705766141917
+
+        # Ratio of mass of daughter particle located at edge of hexagon to that
+        # of its parents mass
+        self.edge_pa_mass_frac = 0.136882287617319
```
prabhuramachandran: Why the term edge?  Why not just daughter?

pysph/sph/swe/basic.py
```diff
@@ -0,0 +1,1677 @@
...
+        return x.copy(), y.copy()
+
+    def _add_edge_pa_prop(self, h0_edge_pa, h_edge_pa, m_edge_pa, x_edge_pa,
+                          y_edge_pa, rho0_edge_pa, rho_edge_pa,
+                          u_prev_step_edge_pa, v_prev_step_edge_pa,
+                          alpha_edge_pa, parent_idx_edge_pa):
+        add_prop = {}
+        for prop in self.props:
+            if prop == 'm':
+                add_prop[prop] = m_edge_pa
```
prabhuramachandran: You can do this with a dictionary and looping instead of this messy and long list of conditions.

pysph/sph/swe/basic.py
```diff
@@ -0,0 +1,1677 @@
...
+    with shock capturing", Int. J. Numer. Meth. Fluids, 69 (2012), pp.
+    1377-1410
+
+    """
+    def __init__(self, dest, sources=None):
+        super(NonDimensionalDensityResidual, self).__init__(dest, sources)
+
+    def post_loop(self, d_psi, d_rho, d_rho_prev_iter, d_idx):
+        # Non-dimensional residual
+        d_psi[d_idx] = abs(d_rho[d_idx]-d_rho_prev_iter[d_idx]) \
```
prabhuramachandran: PEP8, space around operator `-`.

pysph/sph/swe/basic.py
dileepmenon: This class is used in the pre_step method in the example files, wherein an object of ParticleSplit class is instantiated and the method  do_particle_split() is called for particle splitting.

pysph/sph/swe/basic.py
dileepmenon: I have changed this to vertex_pa instead of edge_pa. There are actually daughter particles located at the center of the hexagon and vertices of the hexagon after splitting, so to distinguish between them I used these names.

dileepmenon:
Added the shallow water examples. Only the Tsunami case is left to be added. Have to fix the issue of particle clumping at the inlet boundary which is causing segmentation faults.

prabhuramachandran:
This is superseded by #256, thanks @dileepmenon.  I suspect there are some issues with the fixes I have made to get this working but those can be fixed later.  In any case, I will close this PR and merge #256 once the builds are clean.

PR #109, opened 2018-07-28
Cython Scan

rahulgovind:
The cpy opencl scan does not handle arguments to the segment and output expression correctly. Currently, the scan won't work if the segment or output expression takes an argument which is not also taken by the input expression.

This is the syntax I have in place right now:
```
def input(i, a, b):
    ...

def segment_expr(i, c, d):
    ...

def output(i, ... (reserved keywords) ..., e, f):
    ...

scan = Scan(input=input, output=output, is_segment=segment_expr, ...)
scan(a, b, c, d, e, f)
```

Note the order of the arguments. The input arguments are followed by the segment expression arguments which are then followed by the output arguments. Reserved keywords like item, prev_item need to be inserted before any other arguments for the output function.

I think the following syntax would work better even if it's more verbose because 
1) There is no confusion about the order of arguments 
2) The same argument can be used by both the input and output expressions
```
scan(a=a, b=b, c=c, d=d, e=e, f=f)
```

### DIFF
--- a/pysph/cpy/parallel.py
+++ b/pysph/cpy/parallel.py
@@ -755,7 +755,7 @@ class Scan(object):
         else:
             return False
 
-    def _wrap_ocl_function(self, func):
+    def _wrap_ocl_function(self, func, func_type=None):
         if func is not None:
             self.tp.add(func)
             py_data, c_data = self.cython_gen.get_func_signature(func)
@@ -765,17 +765,19 @@ class Scan(object):
                 func=name,
                 args=', '.join(c_data[1])
             )
-            arguments = convert_to_float_if_needed(
-                ', '.join(c_data[0][1:])
-            )
 
             n_ignore = self._num_ignore_args(c_data)
             arguments = c_data[0][n_ignore:]
             c_args = c_data[1][n_ignore:]
         else:
-            arguments = []
-            expr = None
-            c_args = []
+            if func_type is 'input':
+                arguments = ['__global %(type)s *input' % {'type': self.type}]
+                expr = 'input[i]'
+                c_args = ['input']
+            else:
+                arguments = []
+                expr = None
+                c_args = []
         return expr, arguments, c_args
 
     def _ignore_arg(self, arg_name):
@@ -795,7 +797,7 @@ class Scan(object):
     def _generate(self):
         if self.backend == 'opencl':
             input_expr, input_args, input_c_args = \
-                self._wrap_ocl_function(self.input_func)
+                self._wrap_ocl_function(self.input_func, func_type='input')
 
             output_expr, output_args, output_c_args = \
                 self._wrap_ocl_function(self.output_func)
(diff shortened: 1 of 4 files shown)


pysph/cpy/parallel.py
```diff
@@ -107,6 +106,256 @@
...
+    chunk_new_segment = <int*> malloc(n_thread * stride * sizeof(int))
+    % endif
+
+    % if complex_map:
+    cdef ${type}* map_output
+    map_output = <${type}*> malloc(SIZE * sz)
+    % endif
+
+    if buffer == NULL:
+        abort()
```
prabhuramachandran: I don't like this, we should raise an exception with a memory error or so if we cannot allocate memory and not just abort, that would be a nasty way to exit.

pysph/cpy/parallel.py
```diff
@@ -107,6 +106,256 @@
...
+
+    % if use_segment:
+    free(scan_seg_flags)
+    free(chunk_new_segment)
+    % endif
+
+    % if complex_map:
+    free(map_output)
+    % endif
+
```
prabhuramachandran: I wish it were possible to write this same thing with cpy, I think macros are important and need to think more about this.  Of course, `with` statements won't work and the nogil/parallel will need some work but the rest seems rather straightforward.  Its sad that we need n different implementations of this -- one for each backend.

pysph/cpy/tests/test_parallel.py
```diff
@@ -151,3 +150,169 @@ def output_f(i, ary, item):
...
+        scan = Scan(input_f, output_f, 'a+b', dtype=np.int32, backend=backend,
+                    is_segment=segment_f)
+        scan(a, seg, a)
+        a.pull()
+
+        # Then
+        np.testing.assert_equal(output_actual, a.data)
+
+    def test_segmented_scan_cython(self):
+        self._test_segmented_scan(backend='cython')
```
prabhuramachandran: These all should also be tested with the opencl backend.

pysph/cpy/tests/test_parallel.py
```diff
@@ -151,3 +150,169 @@ def output_f(i, ary, item):
...
+        scan = Scan(output=output_f, scan_expr='a+b',
+                    dtype=np.int32, backend=backend)
+        scan(a, a)
+        a.pull()
+
+        # Then
+        np.testing.assert_equal(expect, a.data)
+
+    def test_scan_last_item_cython_parallel(self):
+        with use_config(use_openmp=True):
```
prabhuramachandran: Is this not tested with opencl?

rahulgovind:
PR is ready for review

prabhuramachandran:
LGTM, could you please update the docs here with a short example and then this PR can be merged. https://github.com/pypr/pysph/blob/master/examples/cpy/README.rst#scan

adityapb:
@prabhuramachandran I think this is RTM.

prabhuramachandran:
Yep, I agree.  Merging.  Thanks @rahulgovind !

PR #112, opened 2018-08-25
GPU support to ParticleArray and run GPU tests on Travis CI

adityapb:


prabhuramachandran:
Thanks!  LGTM, and thanks for adding the pocl testing support.  We need cuda next but this is great for now.

PR #113, opened 2018-09-07
Kernel fix

dineshadepu:
All tests passed.

### DIFF
--- a/pysph/base/kernels.py
+++ b/pysph/base/kernels.py
@@ -206,7 +206,7 @@ class WendlandQuinticC2_1D(object):
 
         return val * fac
 
-    def gradient(self, xij=[0., 0, 0], rij=1.0, h=1.0, grad=[0, 0, 0]):
+    def sigmadwdq(self, rij=1.0, h=1.0):
         h1 = 1. / h
         q = rij * h1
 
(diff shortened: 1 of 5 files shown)


pysph/base/kernels.py
```diff
@@ -89,31 +89,41 @@ def kernel(self, xij=[0., 0, 0], rij=1.0, h=1.0):
 
         return val * fac
 
-    def gradient(self, xij=[0., 0, 0], rij=1.0, h=1.0, grad=[0, 0, 0]):
+    def sigmadwdq(self, rij=1.0, h=1.0):
```
prabhuramachandran: Why the term `sigmadwdq`?  Can't you just call this `dwdq`?

pysph/base/kernels.py
```diff
@@ -89,31 +89,41 @@ def kernel(self, xij=[0., 0, 0], rij=1.0, h=1.0):
...
             else:
-                val = -3.0 * q * (1 - 0.75 * q) * h1 / rij
+                val = -3.0 * q * (1 - 0.75 * q)
         else:
             val = 0.0
 
-        tmp = val * fac
+        return val * fac
+
+    def gradient(self, wdash=1., xij=[0., 0, 0], rij=1.0, h=1.0,
```
prabhuramachandran: Don't do this, as this means the gradient is not computed when you ask for it but requires wdash.  Instead compute wdash inside the gradient.

pysph/base/kernels.py
dineshadepu: I can call dwdq, initially I did that, but the actual formula reads as

```
DWIJ = normalization_factor * dw_dq * (x_ij / (r_ij * h))
```
So keep it clear I named it as `sigmadwdq`

prabhuramachandran: No, this is not standard, in that case return dw_dq.  There is no reference to any formula here.  It would be nice to have a docstring or to update the documentation on this.  There is no standard convention on what "sigma" is.  So please change this.

dineshadepu: Will this get fixed by just adding the documentation, or should I change the name to something else.

In many papers they represent normalization factor by sigma, so I used that notation.

I will change the function name to `dwdq` and write a doc string about what it is.

prabhuramachandran:
Also, did you rebuild c_kernels.pyx, it is not correct that your tests passed.  You must `touch c_kernels.pyx.mako` to rebuild the kernels and only then will the compiled kernels pass the tests (or not).

dineshadepu:
I don't understand touching the c_kernels.pyx. I did `python setup.py develop` twice now, and ran the tests and all tests pass.

dineshadepu:
The Gpu tests fails on travis

prabhuramachandran:
Just run the command `touch c_kernels.pyx.mako` and then run setup.py develop.

prabhuramachandran:
Thanks.  So this exposes a bug with the code generation which is why the tests are failing, I'll fix it soon and then merge this.

PR #118, opened 2018-09-24
Use CPy to rewrite DeviceHelper, move DeviceArray to CPy

adityapb:
This PR does the following things,
1. Moves DeviceArray to CPy and merge it with CPy's Array
2. Use CPy to rewrite DeviceHelper
3. Add tests for Array, DeviceHelper and ParticleArray for all 3 backends


pysph/base/device_helper.py
```diff
@@ -308,40 +180,44 @@ def remove_prop(self, name):
...
-        )
-
-        align_particles_knl(tag_arr, new_indices, num_real_particles,
-                            num_particles)
+        new_indices = array.empty(num_particles, dtype=np.int32,
+                                  backend=self.backend)
+        num_real_particles = array.empty(1, dtype=np.int32,
+                                         backend=self.backend)
+
+        @annotate(i='int', tag_arr='gintp', return_='int')
```
prabhuramachandran: This PR is nice but my only concern here is that each time `align_particles` is called (or any other of the methods using cpy), the function is created again, annotated and the code is generated.  All of the actual kernels should be cached carefully as these functions can be called quite often.

prabhuramachandran: Would be nice to have some sort of abstraction that handles this, as the same pattern is repeated many times.  I guess that since cpy doesn't handle methods, this is a bit harder than it should be but opening up wrapping methods leads to many potential issues like users accessing attributes of the class inside the methods.

pysph/base/particle_array.pyx
```diff
@@ -511,7 +507,7 @@ cdef class ParticleArray:
             the type of particles that need to be removed.
 
         """
-        if self.gpu is not None:
+        if self.gpu is not None and self.backend:
```
prabhuramachandran: Why do you need this `self.backend`?  Can it be false?  Do you mean `self.backend == 'opencl'`?

pysph/base/stratified_sfc_gpu_nnps.pyx
```diff
@@ -102,35 +106,36 @@ cdef class StratifiedSFCGPUNNPS(GPUNNPS):
...
+        radix_sort = cl.algorithm.RadixSort(get_context(),
                 "unsigned int* pids, unsigned long* keys",
                 scan_kernel=GenericScanKernel, key_expr="keys[i]",
                 sort_arg_names=["pids", "keys"])
 
-        (sorted_indices, sorted_keys), evnt = radix_sort(self.pids[pa_index],
-                self.pid_keys[pa_index], key_bits=64)
-        self.pids[pa_index] = sorted_indices
-        self.pid_keys[pa_index] = sorted_keys
+        #FIXME: change key_bits
```
prabhuramachandran: Can you clarify this comment?  What do you mean?

pysph/base/tests/test_particle_array.py
```diff
@@ -34,7 +34,8 @@ class ParticleArrayTest(object):
 
     def test_constructor(self):
         # Default constructor test.
-        p = particle_array.ParticleArray(name='test_particle_array')
+        p = particle_array.ParticleArray(name='test_particle_array',
+                                         backend=self.backend)
```
prabhuramachandran: Is the backend compulsory now?  If so this API is not good, the backend ought to be detected automatically right?  Or is this just for the tests since we change the backend?

pysph/base/particle_array.pyx
```diff
@@ -152,8 +147,8 @@ cdef class ParticleArray:
         # list of output property arrays
         self.output_property_arrays = []
 
-        if get_config().use_opencl:
-            h = ocl.DeviceHelper(self)
+        if self.backend:
```
prabhuramachandran: I don't like this, we should use this: https://github.com/pypr/pysph/blob/master/pysph/cpy/array.py#L6 and perhaps promote that to a place so we can reuse it.  This way the backend can be set once and explicitly if desired instead of have to be passed everywhere?

pysph/base/gpu_nnps_base.pyx
```diff
@@ -422,16 +423,17 @@ cdef class BruteForceNNPS(GPUNNPS):
                 """ % {"data_t" : ("double" if self.use_double else "float")}
 
         brute_force_nbrs = ElementwiseKernel(
```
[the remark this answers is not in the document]
adityapb: Yes

pysph/base/particle_array.pyx
adityapb: This is so that you can use the methods defined in ParticleArray even when `self.gpu` is set. For example see `test_resize_works` in `test_device_helper`

adityapb: Maybe I can have a different value of backend for which this happens, like 'cpu' or something?

PR #120, opened 2018-09-26
Add CUDA scan support to CPy (temporary)

adityapb:
I don't know if there will be any licensing issues with this

### DIFF
--- a/pysph/cpy/transpiler.py
+++ b/pysph/cpy/transpiler.py
@@ -174,7 +174,6 @@ class Transpiler(object):
             __constant__ double pi= 3.141592654f;
             ''')
 
-
     def _handle_symbol(self, name, value):
         backend = self.backend
         value_type = type(value)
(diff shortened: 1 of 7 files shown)


pysph/cpy/transpiler.py
```diff
@@ -123,7 +123,7 @@ def __eq__(self, other):
 
 
 class Transpiler(object):
-    def __init__(self, backend='cython'):
+    def __init__(self, backend='cython', incl_cluda=True):
```
prabhuramachandran: I think we should include cluda all the time so we might as well remove that option.  Do you see a case where we will not want to include the cluda preamble?

adityapb: The GenericScanKernel for both opencl and cuda include CLUDA by default and it adds a prefix to some of the terms to avoid collision with the user's code. By adding it again, we are redefining some macros, opencl works fine but nvcc gives and error

adityapb: No, I mean we're including the cluda preamble when GenericScanKernel already does it by default (see [this](https://github.com/adityapb/pysph/blob/0622d1b37f8f96975aee5917ea1fd099688e594d/pysph/cpy/cuda.py#L95)). So it ends up being added twice and so we accidentally redefine some macros which gives a compilation error with nvcc

prabhuramachandran: Got it.

PR #123, opened 2018-10-02
Add jit module to CPy

adityapb:
This module allows you to use map, reduce and scan without adding type annotations.
I don't know if you wanted this @prabhuramachandran , but I got a little carried away by the idea.

### DIFF
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -1,3 +1,8 @@
+"""Shows the use of annotate without any type information.
+The type information is extracted from the arguments passed
+and the function is annotated and compiled at runtime.
+"""
+
 from pysph.cpy.api import annotate, Elementwise, wrap, get_config
 import numpy as np
 
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -3,7 +3,9 @@ from pysph.cpy.jit import ElementwiseJIT
 import numpy as np
 
 def axpb(i, x, y, a, b):
-    y[i] = a*sin(x[i]) + b
+    xi = declare('double')
+    xi = x[i]
+    y[i] = a*sin(xi) + b
 
 x = np.linspace(0, 1, 10000)
 y = np.zeros_like(x)
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -1,7 +1,8 @@
 from pysph.cpy.api import Elementwise, annotate, wrap, get_config
-from pysph.cpy.jit import ElementwiseJIT
+from pysph.cpy.jit import jit, ElementwiseJIT
 import numpy as np
 
+@jit
 def axpb(i, x, y, a, b):
     xi = declare('double')
     xi = x[i]
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -1,5 +1,4 @@
-from pysph.cpy.api import Elementwise, wrap, get_config
-from pysph.cpy.jit import annotate
+from pysph.cpy.api import annotate, Elementwise, wrap, get_config
 import numpy as np
 
 
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -1,9 +1,9 @@
-from pysph.cpy.api import Elementwise, annotate, wrap, get_config
-from pysph.cpy.jit import jit, ElementwiseJIT
+from pysph.cpy.api import Elementwise, wrap, get_config
+from pysph.cpy.jit import annotate
 import numpy as np
 
 
-@jit
+@annotate
 def axpb(i, x, y, a, b):
     xi = declare('double')
     xi = x[i]
@@ -18,5 +18,5 @@ b = 3.0
 backend = 'opencl'
 get_config().use_openmp = True
 x, y = wrap(x, y, backend=backend)
-e = ElementwiseJIT(axpb, backend=backend)
+e = Elementwise(axpb, backend=backend)
 e(x, y, a, b)
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -2,11 +2,13 @@ from pysph.cpy.api import Elementwise, annotate, wrap, get_config
 from pysph.cpy.jit import jit, ElementwiseJIT
 import numpy as np
 
+
 @jit
 def axpb(i, x, y, a, b):
     xi = declare('double')
     xi = x[i]
-    y[i] = a*sin(xi) + b
+    y[i] = a * sin(xi) + b
+
 
 x = np.linspace(0, 1, 10000)
 y = np.zeros_like(x)
--- a/examples/cpy/axpb_jit.py
+++ b/examples/cpy/axpb_jit.py
@@ -1,0 +1,17 @@
+from pysph.cpy.api import Elementwise, annotate, wrap, get_config
+from pysph.cpy.jit import ElementwiseJIT
+import numpy as np
+
+def axpb(i, x, y, a, b):
+    y[i] = a*sin(x[i]) + b
+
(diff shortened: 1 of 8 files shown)


pysph/cpy/transpiler.py
```diff
@@ -158,6 +158,10 @@ def __init__(self, backend='cython', incl_cluda=True):
             self.header = cluda + dedent('''
             #define max(x, y) fmax((double)(x), (double)(y))
 
+            #ifdef __APPLE__
+            #define M_PI 3.141592654f
```
prabhuramachandran: Too low precision.

adityapb:
The `@jit` decorator is only to tell the `AnnotationHelper` whether to visit the function to find it's return type. For example if you're doing something like `return sin(x)`, it wouldn't visit `sin` but would default to `double`. I guess we could use something like `@mark` instead?
Also yes, I think it should do the jitting automatically when the function has no annotations.

pysph/cpy/parallel.py
```diff
@@ -507,7 +507,7 @@ def __call__(self, *args, **kw):
 
 class Elementwise(object):
     def __init__(self, func, backend='cython'):
-        if hasattr(func, '__annotations__') and getattr(func, '__annotations__'):
+        if hasattr(func, 'has_annotations'):
```
prabhuramachandran: Instead of doing this, you can do this:
```
getattr(func, '__annotations__', None)
```
This avoids having to set another attribute and rely on that.

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+        if not hasattr(obj, 'cache'):
+            obj.cache = dict()
+        key = tuple([get_ctype_from_arg(arg) for arg in args])
+        if key not in obj.cache:
+            obj.cache[key] = f(obj, *args, **kwargs)
+        return obj.cache[key]
+    return wrapper
+
+
+def getargspec(f):
```
prabhuramachandran: Another way to do this instead is:
`getargspec = getattr(inspect, 'getfullargspec', getattr(inspect, 'getargspec'))`

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+        if isinstance(node.func, ast.Name) and \
+                node.func.id not in BUILTINS:
+            if f is None or isinstance(f, Extern):
+                return None
+            else:
+                arg_types = []
+                for arg in node.args:
+                    arg_type = self.visit(arg)
+                    if not arg_type:
+                        msg = "Function called is not marked by the jit "\
```
prabhuramachandran: Use a triple quoted string with `textwrap.dedent` put brackets to avoid the `\` at the end of the line?

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+        return_type = 'u%s' % return_type
+    return return_type
+
+
+class AnnotationHelper(ast.NodeVisitor):
+    def __init__(self, func, arg_types):
+        self.func = func
+        self.arg_types = arg_types
+        self.var_types = arg_types.copy()
+        self.children = {}
```
prabhuramachandran: Can you add a comment on what "children" means here or perhaps a better variable name?

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+        if isinstance(x, array.Array):
+            return x.dev
+        elif isinstance(x, np.ndarray):
+            return x
+        else:
+            return np.asarray(x)
+
+    def __call__(self, *args, **kw):
+        c_func = self._generate_kernel(*args)
+        c_args = [self._massage_arg(x) for x in args]
```
prabhuramachandran: This is something for later and not this PR but if we are going to call the same jitted function repeatedly, we should find a way to not have to keep massaging the args each time.

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+            event.record()
+            event.synchronize()
+            return result.get()
+
+
+class ScanJIT(parallel.ScanBase):
+    def __init__(self, input=None, output=None, scan_expr="a+b",
+                 is_segment=None, dtype=np.float64, neutral='0',
+                 complex_map=False, backend='opencl'):
+        backend = array.get_backend(backend)
```
prabhuramachandran: This is rather similar to the ReductionJIT so should one be a subclass?

prabhuramachandran: Note, I know that a lot of the earlier code I wrote requires a massive refactor as that was my first cut... :)

pysph/cpy/tests/test_jit.py
```diff
@@ -0,0 +1,636 @@
...
+def g(x):
+    return x
+
+
+@annotate
+def h(a, b):
+    return g(a) * g(b)
+
+
+class TestAnnotationHelper(unittest.TestCase):
```
prabhuramachandran: Am really glad you've added a lot of tests and are using the Given/When/Then style!  If possible, make sure you have 100% coverage of the annotation helper and other code.  This will be easier when we pull out cpy from pysph but if you can, please measure coverage and make sure it covers all the code.  If this is a pain, we should do this once we pull cpy out.

pysph/cpy/types.py
```diff
@@ -236,6 +238,7 @@ def f(i, x, y):
 
     def wrapper(func):
         func.__annotations__ = data
+        func.has_annotations = True
```
prabhuramachandran: I think this can be avoided as mentioned above.

examples/cpy/axpb_jit.py
```diff
@@ -0,0 +1,22 @@
+from pysph.cpy.api import Elementwise, wrap, get_config
+from pysph.cpy.jit import annotate
+import numpy as np
+
+
```
prabhuramachandran: Please add a helpful comment here, explaining this a bit.

examples/cpy/axpb_jit.py
```diff
@@ -0,0 +1,22 @@
+from pysph.cpy.api import Elementwise, wrap, get_config
+from pysph.cpy.jit import annotate
```
prabhuramachandran: I don't really like that you are defining another analogous annotate, there should be one `annotate` decorator from api and not one from jit and another from types.  This will make it hard for users and developers.  Basically if there are arguments, it will treat it as the old annotate and if there are none, it jits.  So basically merge the two into one.

examples/cpy/vm_elementwise_jit.py
```diff
@@ -0,0 +1,81 @@
+import numpy as np
+from math import pi
+import time
+
+from pysph.cpy.config import get_config
+from pysph.cpy.api import declare, annotate
+from pysph.cpy.parallel import Elementwise
+from pysph.cpy.array import wrap
+from pysph.cpy.jit import annotate
```
prabhuramachandran: Again, one unique annotate which jits if needed.

pysph/cpy/jit.py
```diff
@@ -0,0 +1,523 @@
...
+        func.is_jit = True
+        return func
+
+    if func is None:
+        return wrapper
+    else:
+        return wrapper(func)
+
+
+def memoize(f):
```
prabhuramachandran: Can't we use pytools?  Is that not a hard dependency already?  If it is not, I understand but no point reproducing this no?

pysph/cpy/jit.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def kernel_cache_key_kwargs(obj, **kwargs):
+    key = [get_ctype_from_arg(arg) for arg in kwargs.values()]
+    key.append(obj)
+    return tuple(key)
+
+
+def getargspec(f):
+    getargspec_f = getattr(inspect, 'getfullargspec',
```
prabhuramachandran: Instead of doing this each time, you can define this at the module level so the getattr is done once, but this is a minor thing.  Can be done later.

prabhuramachandran:
LGTM, thanks!

PR #126, opened 2018-10-17
Add local memory CUDA support to Kernel

adityapb:


pysph/cpy/translator.py
```diff
@@ -532,6 +539,26 @@ def visit_FunctionDef(self, node):
...
+                                  '(%(dtype)s*) shared_buff;')
+                    local_decl = local_decl % {'dtype': dtype, 'arg': arg}
+                    decls.append(local_decl)
+                    prev_arg = arg
+                else:
+                    local_decl = ('%(dtype)s* %(arg)s = (%(dtype)s*) '
+                                  '&%(prev_arg)s[size_%(prev_arg)s];')
+                    local_decl = local_decl % {'dtype': dtype, 'arg': arg,
+                                               'prev_arg': prev_arg}
+                    decls.append(local_decl)
```
prabhuramachandran: This whole block of code is not at all intentional and it is very difficult to see what you are trying to do.  I think it really needs to be refactored.

adityapb:
This is interesting. I've updated my branch, but the PR doesn't seem to be updated.

adityapb:
This is interesting. I've updated my branch, but the PR doesn't seem to be updated.

pysph/cpy/translator.py
```diff
@@ -693,3 +694,141 @@ def __init__(self, detect_type=ocl_detect_type, known_types=None):
 
     def _get_self_type(self):
         return KnownType('GLOBAL_MEM %s*' % self._class_name)
+
+
+class CUDAConverter(CConverter):
```
prabhuramachandran: Please try to reuse as much code as possible, derive this from `OpenCLConverter` so we don't have to update things in two places.

pysph/cpy/translator.py
```diff
@@ -693,3 +694,141 @@ def __init__(self, detect_type=ocl_detect_type, known_types=None):
...
+            'GID_0', 'GID_1', 'GID_2',
+            'LDIM_0', 'LDIM_1', 'LDIM_2',
+            'GDIM_0', 'GDIM_1', 'GDIM_2'
+        ))
+        self._local_decl = None
+
+    def _get_self_type(self):
+        return KnownType('GLOBAL_MEM %s*' % self._class_name)
+
+    def _get_function_args(self, node):
```
prabhuramachandran: Again, please try to refactor to minimize cut/pasting.

pysph/cpy/translator.py
```diff
@@ -693,3 +694,141 @@ def __init__(self, detect_type=ocl_detect_type, known_types=None):
...
+        fname = obj.__name__
+        self._annotations[fname] = getattr(obj, '__annotations__', None)
+        self._local_decl = self._get_local_info(obj)
+        code = self.convert(src)
+        self._local_decl = None
+        self._annotations = {}
+        return code
+
+    def visit_FunctionDef(self, node):
+        assert node.args.vararg is None, \
```
prabhuramachandran: Again a lot of similar code that is cut pasted.  Any cut pasting makes it very hard to update things later.  I don't mind adding an empty method to the base class that is only used for CUDA but you need to think about reducing cut/pasting and try to avoid sending PRs with any of it. :) Also please measure coverage for just the translator alone and ensure that the translator has almost 100% coverage.

pysph/cpy/translator.py
```diff
@@ -693,3 +703,57 @@ def __init__(self, detect_type=ocl_detect_type, known_types=None):
...
+            for arg, dtype in self._local_decl.items():
+                if len(decls) == 1:
+                    local_decl = ('%(dtype)s* %(arg)s = '
+                                  '(%(dtype)s*) shared_buff;')
+                    local_decl = local_decl % {'dtype': dtype, 'arg': arg}
+                    decls.append(local_decl)
+                    prev_arg = arg
+                else:
+                    local_decl = ('%(dtype)s* %(arg)s = (%(dtype)s*) '
+                                  '&%(prev_arg)s[size_%(prev_arg)s];')
```
prabhuramachandran: There is one big problem here if the sizes are not aligned to 4 bytes as discussed in the cuda programming guide.  So this should perhaps be carefully documented or perhaps a warning given if there is an alignment problem.

prabhuramachandran:
I've filed an issue #130, for you to address.  I will merge this for now.  Thanks!

PR #127, opened 2018-10-24
Use new strides for kernel corrections.

prabhuramachandran:
Using constants for this is fragile as if we underestimate the size
(especially for the taylor green problem) the code will segfault.  This
is now fixed cleanly.


prabhuramachandran:
@adityapb -- I guess this will need to be merged after #126 as it fixes the travis issues.

adityapb:
I'll update PR #126 today afternoon

PR #138, opened 2018-11-16
Warn user if h or m of a particle is zero.

dineshadepu:


pysph/solver/application.py
```diff
@@ -10,6 +10,8 @@
 import socket
 import sys
 import time
+import numpy
```
prabhuramachandran: `import numpy as np`?

pysph/solver/application.py
```diff
@@ -808,6 +810,16 @@ def _create_particles(self, particle_factory, *args, **kw):
             else:
                 self.particles = particle_factory(*args, **kw)
 
+            for pa in self.particles:
+                if numpy.all(pa.h) == 0.0:
```
prabhuramachandran: how about `np.max(pa.h) < 1e-12` ?

pysph/solver/application.py
```diff
@@ -808,6 +810,16 @@ def _create_particles(self, particle_factory, *args, **kw):
...
+                    warnings.warn(
+                        "'h' for particle array '{}' is 0.0".format(pa.name),
+                        UserWarning)
+                if numpy.all(pa.m == 0.0):
```
prabhuramachandran: Similarly here?

prabhuramachandran:
So the reason the test fails is that sometimes particle arrays can be empty -- like in an outlet that is initially empty.  Checking for a maximum element on an empty array will cause an error.  So you need to add a check for the array in that it has at least 1 particle.  Run the example that is failing to see the error.

prabhuramachandran:
Thanks.

PR #139, opened 2018-11-21
add __repr__ to MultiStageEquations

abhinavmuta:
"eqn_info" is set using a list variable, so for multi stage equations this needs to be changed.

### DIFF
--- a/pysph/sph/equation.py
+++ b/pysph/sph/equation.py
@@ -348,7 +348,9 @@ def get_arrays_used_in_equation(equation):
     """
     src_arrays = set()
     dest_arrays = set()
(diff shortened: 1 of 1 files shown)


pysph/solver/solver.py
```diff
@@ -198,6 +199,9 @@ def setup(self, particles, equations, nnps, kernel=None, fixed_h=False):
         )
 
         sep = '-'*70
+
+        if isinstance(equations, MultiStageEquations):
```
prabhuramachandran: A better approach would be to overload `__repr__` on the MultiStageEquations class as is done for the Group and Equation classes.

abhinavmuta: Thanks! did the changes.

PR #141, opened 2018-11-24
clean-up logging code, include kernel and particle info.

abhinavmuta:
The MultiStageEquation `__iter__` method may not be clean, any suggestions?

### DIFF
--- a/pysph/solver/application.py
+++ b/pysph/solver/application.py
@@ -1069,6 +1069,7 @@ class Application(object):
             nnps=nnps,
             kernel=kernel,
             fixed_h=fixed_h)
+        self._log_solver_info(solver)
 
         # add solver interfaces
         self.command_manager = CommandManager(solver, self.comm)
@@ -1217,15 +1218,15 @@ class Application(object):
             self.solver.add_post_step_callback(obj.post_step)
 
     def _message(self, msg):
-        if self.options.quiet:
-            return
         if self.num_procs == 1:
             logger.info(msg)
-            print(msg)
+            if not self.options.quiet:
+                print(msg)
         elif (self.num_procs > 1 and self.rank in (0, 1)):
             s = "Rank %d: %s" % (self.rank, msg)
             logger.info(s)
-            print(s)
+            if not self.options.quiet:
+                print(s)
 
     def _write_info(self, filename, **kw):
         """Write the information dictionary to given filename. Any extra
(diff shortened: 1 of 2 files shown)


pysph/solver/application.py
```diff
@@ -1236,6 +1237,24 @@ def _write_info(self, filename, **kw):
...
+        kernel_info = '%s(dim=%s)' % (kernel_name, solver.dim)
+
+        logger.info(
+            'No of particles:\n%s\n  %s\n%s' % (sep, particle_info, sep)
+        )
+        logger.info('Using kernel:\n%s\n  %s\n%s' % (sep, kernel_info, sep))
+        logger.info(
+            'Using integrator:\n%s\n  %s\n%s' % (sep, solver.integrator, sep)
+        )
+        logger.info('Using equations:\n%s\n%s\n%s' % (sep, eqn_info, sep))
```
42B: Is there any reason why these `logger` calls aren't passing in the variables as shown in the Python docs?

```python
logger.info('Using equations:\n%s\n%s\n%s', sep, eqn_info, sep)  # ?
```

pysph/sph/equation.py
```diff
@@ -871,3 +871,15 @@ def __repr__(self):
         return '%s(groups=[[\n%s\n    ]]\n)' % (
             name, '], [\n'.join(kw)
         )
+
+    def __iter__(self):
```
prabhuramachandran: Don't you just need to do return `iter(self.groups)` here and not have to define a `__next__`

pysph/sph/equation.py
```diff
@@ -871,3 +871,15 @@ def __repr__(self):
...
+        self._cntr = 1
+        return self
+
+    def __next__(self):
+        # import pudb; pudb.set_trace()
```
prabhuramachandran: You can remove this.

pysph/solver/application.py
```diff
@@ -1236,6 +1237,24 @@ def _write_info(self, filename, **kw):
...
+    def _log_solver_info(self, solver):
+        sep = '-'*70
+        eqn_info = '[\n' + ',\n'.join([str(e) for e in self.equations]) + '\n]'
+        particle_info = '\n  '.join(['%s: %d' % (p.name, len(p.gid)) for p in
+                                    solver.particles])
+
+        kernel_name = solver.kernel.__class__.__name__
+        kernel_info = '%s(dim=%s)' % (kernel_name, solver.dim)
+
+        logger.info(
```
prabhuramachandran: Can use `self._message` for the particle information alone as it is often useful to see.  Also please provide information on the NNPS object used.

pysph/solver/application.py
abhinavmuta: Oh didn't know that, thanks!

pysph/sph/equation.py
abhinavmuta: removed it, now we check if the `equation` is a list or not in the logging method, thanks!

pysph/solver/application.py
abhinavmuta: updated with nnps info, thanks!

pysph/solver/application.py
42B: Sorry, I should have included a link to the relevant docs.

* https://docs.python.org/3/library/logging.html#logging.debug

PR #143, opened 2018-11-26
Add Generalized TVF scheme

abhinavmuta:
For now the scheme works for only fluid
problems like Taylor-Green, but extending to
fluid-solid problems should be straight forward.

Taylor-Green (TG) problem is add to "old_examples/", so that TG example in "examples/" is not cluttered.


pysph/sph/wc/gtvf.py
```diff
@@ -0,0 +1,655 @@
...
+                omegaT[d*j + i] = omega[d*i + j]
+
+        tmp1, tmp2 = declare('matrix(9)', 2)
+        mat_mult(sigmai, omegaT, d, tmp1)
+        mat_mult(omega, sigmai, d, tmp2)
+
+        for i in range(d):
+            for j in range(d):
+                d_asigma[d_idx*d*d + d*i + j] = 2*G * eps[d*i + j]
+                d_asigma[d_idx*d*d + d*i + j] += -2*G * eps_trace/3.0
```
prabhuramachandran: The trace term should be subtracted only on the diagonals and not for every i and j.

pysph/sph/wc/gtvf.py
```diff
@@ -0,0 +1,655 @@
...
+        self.solver = Solver(
+            dim=self.dim, integrator=integrator, kernel=kernel,
+            output_at_times=[0, 0.2, 0.4, 0.8], **kw
+        )
+
+    def get_equations(self):
+        from pysph.sph.wc.transport_velocity import StateEquation
+        sources = self.fluids
+
+        eq1, stage1 = [], []
```
prabhuramachandran: So this also does not incorporate solids so we should either not allow users to pass them in the constructor or add support for it.

pysph/sph/wc/gtvf.py
abhinavmuta: thanks, fixed it!

pysph/sph/wc/gtvf.py
abhinavmuta: removed it from the scheme for now.

pysph/sph/wc/gtvf.py
```diff
@@ -0,0 +1,655 @@
...
+        """
+        self.G = G
+        self.dim = dim
+        super(DeviatoricStressRate, self).__init__(dest, sources)
+
+    def initialize(self, d_idx, d_asigma):
+        i = declare('int')
+        for i in range(9):
+            d_asigma[d_idx*9 + i] = 0.0
+            printf("%d, %f\n", d_idx*9 + i, d_asigma[d_idx*9 + 1])
```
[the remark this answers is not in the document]
abhinavmuta: thanks!

pysph/sph/wc/gtvf.py
```diff
@@ -0,0 +1,655 @@
...
+             DWIJ):
+        d, i = declare('int', 3)
+        sigmaij = declare('matrix(9)')
+        res = declare('matrix(3)')
+
+        d = self.dim
+
+        for i in range(d):
+            sigmaij[i] = d_sigma[d_idx*9 + i] + s_sigma[s_idx*9 + i]
+            sigmaij[i] = d_sigma[d_idx*9 + i] + s_sigma[s_idx*9 + i]
```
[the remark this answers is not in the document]
abhinavmuta: corrected it, thanks!

PR #145, opened 2018-11-28
preserve stride info in '_ensure_properties'

abhinavmuta:
When using 'setup_properties' method in a scheme,
pass the dummy.stride value to '_ensure_properties'
to preserve the info.

### DIFF
--- a/pysph/sph/scheme.py
+++ b/pysph/sph/scheme.py
@@ -108,19 +108,27 @@ class Scheme(object):
         pa : ParticleArray
             Desired particle array.
         desired_props : sequence
-            Desired properties to have in the array.
+            Desired properties to have in the array, can be a list of strings
+            or dicts with stride info or both.
         clean : bool
             Remove undesirable properties.
         """
-        all_props = set(desired_props)
+        all_props = {}
+        for p in desired_props:
+            if isinstance(p, dict):
+                all_props.update({p['name']: p})
+            elif p not in all_props:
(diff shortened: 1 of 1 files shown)


pysph/sph/scheme.py
```diff
@@ -97,7 +97,8 @@ def setup_properties(self, particles, clean=True):
 
     # Private protocol ###################################################
 
-    def _ensure_properties(self, pa, desired_props, clean=True):
+    def _ensure_properties(self, pa, desired_props, stride_info=None,
```
prabhuramachandran: put the stride_info at the end, or you may have to track all uses of `_ensure_properties` where clean is set using positional arguments.

pysph/sph/scheme.py
```diff
@@ -120,7 +123,11 @@ def _ensure_properties(self, pa, desired_props, clean=True):
 
         to_add = all_props - set(pa.properties.keys())
         for prop in to_add:
-            pa.add_property(prop)
+            if stride_info is not None and prop in stride_info:
```
prabhuramachandran: Maybe it would be better to pass in the desired_props as a sequence or strings or dicts.  If its a dict, you use it as `add_property(**item)` else as just `add_property(item)`.

abhinavmuta: changed it to dict, also added few lines which ensure that the desired_props have no duplicates even when the user does the wrong thing.

PR #153, opened 2018-12-16
Add particle position shift.

abhinavmuta:
This is added to maintain better uniform
distribution of particles.

### DIFF
--- a/pysph/sph/wc/shift.py
+++ b/pysph/sph/wc/shift.py
@@ -1,25 +1,55 @@
+"""
+Shift particle positions
+########################
+
+Equations to maintain uniform particle distribution.
+
+There are two ways proposed:
+1. 'SimpleShift' is the earlier one, see [XuStaLau2009].
+2. 'FickianShift' is based on Fick's diffusion law, see [LiXuStaRo2012].
+
+TODO: Implement for free surface.
+
+References
+----------
+
+    .. [XuStaLau2009] Rui Xu, Peter Stansby, Dominique Laurence (2009)
+        Accuracy and stability in incompressible SPH (ISPH) based
+        on the projection method and a new approach.
+
+    .. [LiXuStaRo2012] S.J Lind, R. Xu, P.K. Stansby, B.D. Rogers (2012)
+        Incompressible smoothed particle hydrodynamics for free-surface flows:
+        A generalised diffusion-based algorithm for stability and validations
+        for impulsive flows and propagating waves.
+
+    .. [SkLiStaRo2013] Alex Skillen, S. Lind, P.K. Stansby, B.D. Rogers (2013)
+        Incompressible smoothed particle hydrodynamics (SPH) with reduced
+        temporal noise and generalised Fickian smoothing applied to
+        body water slam and efficient wave body interaction.
+"""
+
 from math import sqrt
 from compyle.api import declare
 from pysph.sph.equation import Equation
-from pysph.base.reduce_array import serial_reduce_array
+from pysph.base.reduce_array import parallel_reduce_array
 from pysph.solver.tools import Tool
 
 
 class SimpleShift(Equation):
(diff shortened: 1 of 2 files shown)


pysph/sph/wc/shift.py
```diff
@@ -0,0 +1,226 @@
...
+
+class SimpleShift(Equation):
+    """See the paper by R. Xu et al. (JCP 2009), equation(35)
+    """
+    def __init__(self, dest, sources, const=0.04):
+        self.beta = const
+        super(SimpleShift, self).__init__(dest, sources)
+
+    def py_initialize(self, dst, t, dt):
+        from math import sqrt
```
prabhuramachandran: Is this import needed? and shouldn't it be from `numpy import sqrt`?

pysph/sph/wc/shift.py
```diff
@@ -0,0 +1,226 @@
...
+        vmag = dst.u[:]*dst.u[:] + dst.v[:]*dst.v[:] + dst.w[:]*dst.w[:]
```
prabhuramachandran: Can't this be simply `vmag = dst.u**2 + dst.v **2 + dst.w**2`

pysph/sph/wc/shift.py
```diff
@@ -0,0 +1,226 @@
...
+        vmax = serial_reduce_array(vmag, 'max')
```
prabhuramachandran: Technically this should be a `parallel_reduce_array` but it is fine for this use case I imagine.

pysph/sph/wc/shift.py
```diff
@@ -0,0 +1,226 @@
...
+        d_dpos[d_idx*3] = fac*grad_c[0]
+        d_dpos[d_idx*3 + 1] = fac*grad_c[1]
+        d_dpos[d_idx*3 + 2] = fac*grad_c[2]
+
+        d_x[d_idx] += d_dpos[d_idx*3]
+        d_y[d_idx] += d_dpos[d_idx*3 + 1]
+        d_z[d_idx] += d_dpos[d_idx*3 + 2]
+
+
+class CorrectVelocities(Equation):
```
prabhuramachandran: Docstring for this?

pysph/sph/wc/shift.py
```diff
@@ -0,0 +1,226 @@
...
+        d_x[d_idx] += d_dpos[d_idx*3]
+        d_y[d_idx] += d_dpos[d_idx*3 + 1]
+        d_z[d_idx] += d_dpos[d_idx*3 + 2]
+
+
+class FickianShift(Equation):
+    """See the paper by S. J. Lind et al. (JCP 2012), equation(21-24),
+    for the constant see A. Skillen et al. (CMAME 2013), equation(13).
+    """
+    def __init__(self, dest, sources, const=10, tensile_const=0.2,
```
prabhuramachandran: Why the name `const`, can there be a better name that is similar to the one used in the paper?

pysph/sph/wc/shift.py
abhinavmuta: fixed, thanks!

pysph/sph/wc/shift.py
abhinavmuta: changed it, thanks.

pysph/sph/wc/shift.py
abhinavmuta: added doc strings.

pysph/sph/wc/shift.py
abhinavmuta: the papers names it as 'C', now changed it to 'fickian_const', should I change it ?

pysph/sph/wc/shift.py
abhinavmuta: this is not needed! I tried to not import numpy at all, but now changed it to a numpy.sqrt operation as it is much clear.

prabhuramachandran:
LGTM, thanks!

PR #170, opened 2019-02-14
Fix Gas Dynamics issues

ananyo-work:
-  fix examples for Kunal (2014)
-  fix mpm not using adaptive_h_scheme and add wallboundary 
-  fix riemann_solver.solve not accounting for initial discontinuity position

### DIFF
--- a/pysph/examples/gas_dynamics/shocktube.py
+++ b/pysph/examples/gas_dynamics/shocktube.py
@@ -172,11 +172,10 @@ class ShockTube2D(Application):
         print(gamma)
         riemann_solver.set_gamma(gamma)
         rho_e, u_e, p_e, e_e, x_e = riemann_solver.solve(
-            x_min=-0.5, x_max=0.5, x_0=0,
+            x_min=0, x_max=1, x_0=0.5,
             t=self.tf, p_l=self.pl, p_r=self.pr, rho_l=self.rhol,
             rho_r=self.rhor, u_l=self.ul, u_r=self.ur, N=101
         )
-        x_e = x_e + self.x0
 
         x = pa.x
         u = pa.u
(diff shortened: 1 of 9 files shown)


pysph/examples/gas_dynamics/shocktube.py
```diff
@@ -55,42 +76,169 @@ def create_particles(self):
...
+    def post_process(self):
+        try:
+            from matplotlib import pyplot
+        except ImportError:
+            print("Post processing requires matplotlib.")
+            return
+        import os
+        from pysph.solver.utils import load
+        from pysph.examples.gas_dynamics import riemann_solver
+        outfile = self.output_files[-1]
```
prabhuramachandran: You need to guard for this, as the tests do not generate any output, so if `len(self.output_files)` is 0 then you should just return.

ananyo-work:
> Hello @ananyo-work! Thanks for updating the PR.
> 
>     * In the file [`pysph/examples/gas_dynamics/riemann_solver.py`](https://github.com/pypr/pysph/blob/8309489509cf18a1ab194d0b0c0f1663eb8ed354/pysph/examples/gas_dynamics/riemann_solver.py), following are the PEP8 issues :
> 
> 
> > [Line 130:5](https://github.com/pypr/pysph/blob/8309489509cf18a1ab194d0b0c0f1663eb8ed354/pysph/examples/gas_dynamics/riemann_solver.py#L130): [E731](https://duckduckgo.com/?q=pep8%20E731) do not assign a lambda expression, use a def
> 
> ##### Comment last updated on February 22, 2019 at 10:43 Hours UTC

So, the code has more lambdas, on lines 96 and 98. These have been here even before I edited the riemann_solver. Should I just leave them there?

prabhuramachandran:
Well, you can certainly clean them up and remove the lambdas.

prabhuramachandran:
I will merge this for now.  Thanks!

PR #171, opened 2019-02-15
Oscillating plate example using standard SPH

dineshadepu:


pysph/examples/solid_mech/oscillating_plate_gray.py
```diff
@@ -0,0 +1,250 @@
...
+        integrator = EPECIntegrator(plate=SolidMechStep())
+
+        dt = 1e-6
+        tf = 1
+
+        solver = Solver(kernel=kernel, dim=2, integrator=integrator, dt=dt,
+                        tf=tf, adaptive_timestep=False)
+        return solver
+
+    def create_equations(self):
```
prabhuramachandran: Would it be possible to make this into a simple scheme perhaps so we don't have to keep rewriting the equations?  Also can the example name be simply `oscillating_plate.py` so when more schemes are available, we can easily switch with the same example.

PR #175, opened 2019-02-25
Fixed the 3D viewer, added the 'save all plots' option to the 2D viewer

thisisrohan:
Debugged and updated the 3D viewer, added the 'save all plots' option to the 2D viewer.

### DIFF
--- a/pysph/tools/ipy_viewer.py
+++ b/pysph/tools/ipy_viewer.py
@@ -888,7 +888,7 @@ class Viewer3D(Viewer):
                 scatters.z = data[array_name].z,
                 scatters.color = c
                 pa_widgets = self._widgets.particles[array_name]
-            if pa_widgets.velocity_vectors.value == True:
+            if pa_widgets.velocity_vectors.value is True:
                 vectors = self.vectors[array_name]
                 vectors.x = data[array_name].x
                 vectors.y = data[array_name].y
@@ -998,6 +998,6 @@ class Viewer3D(Viewer):
                         boundaries=ticks,
                     )
                     self.cbars[array_name].set_label(
-                                    array_name + " : " + pa_widgets.scalar.value
+                                    array_name+" : "+pa_widgets.scalar.value
                             )
             display(self.pltfigure)
--- a/pysph/tools/ipy_viewer.py
+++ b/pysph/tools/ipy_viewer.py
@@ -18,7 +18,6 @@ class Viewer(object):
 
         self.path = path
         self.paths_list = get_files(path)
-        self._delta_t = -1
 
         # Caching #
         # Note : Caching is only used by get_frame and widget handlers.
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Also, what version of ipyvolume does this work with?

pysph/tools/ipy_viewer.py
```diff
@@ -349,7 +358,7 @@ def _configure_plot(self):
         '''
         Set attributes for plotting.
         '''
-
+        plt.ion()
```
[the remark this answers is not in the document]
thisisrohan: So ipyvolume does not have an inherent legend attribute like matplotlib, and to display the legend for the `3DViewer`, we're plotting it separately in a matplotlib 2D figure that is then being displayed alongside the 3D plot. To get a static plot for the legned, `plt.ioff()` is being used. I thus included `plt.ion()` in  `2DViewer` to make sure we get an interactive figure when the `interactive_plot()` method is called on an instance of  `2DViewer`.
The use case is someone calling `interactive_plot()` on a `2DViewer` instance after calling it on a `3DViewer` instance.

thisisrohan:
This works on ipyvolume 0.5.1

pysph/tools/ipy_viewer.py
```diff
@@ -1,11 +1,13 @@
...
-from pysph.solver.utils import load, get_files
-from IPython.display import display, Image
+import os
+from pysph.solver.utils import load, get_files, mkdir
+from IPython.display import display, Image, clear_output
 import ipywidgets as widgets
 import matplotlib.pyplot as plt
 import matplotlib as mpl
 import numpy as np
+import ipyvolume.pylab as p3
```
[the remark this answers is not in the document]
thisisrohan: Updated the PR, now it only gets imported if someone calls the `interactive_plot()` method on a `3DViewer` instance.

pysph/tools/ipy_viewer.py
```diff
@@ -656,8 +676,26 @@ def _save_figure_handler(self, change):
...
+        if change['new']:
+            self._widgets.save_figure.disabled = True
+        elif not change['new']:
+            self._widgets.save_figure.disabled = False
+
+    def _save_plot(self):
+
+        abs_path = os.path.abspath(self.path)
+        if self._widgets.save_all_plots.value:
+            mkdir('/all_plots')
```
prabhuramachandran: This should really be `mkdir('all_plots')`!

prabhuramachandran:
I am not too fond of the import in the plot function currently but this does look pretty good, except for one tiny error I spotted, please fix that, I've left a comment.

pysph/tools/ipy_viewer.py
```diff
@@ -656,8 +676,26 @@ def _save_figure_handler(self, change):
...
+            self.figure.savefig(
+                'frame_%s.png' % self._widgets.frame.value,
```
prabhuramachandran: This should be `all_plots/frame...` right?

pysph/tools/ipy_viewer.py
```diff
@@ -683,9 +721,9 @@ def __init__(self, particlearray):
         self.scalar_cmap.owner = self.array_name
         self.velocity_vectors = widgets.Checkbox(
             value=False,
-            description="Vectors",
+            description="Velocity Vectors",
             disabled=False,
-            layout=widgets.Layout(width='100px', display='flex')
+            layout=widgets.Layout(width='300px', display='flex')
```
prabhuramachandran: Do we need fixed sizes here?  Does it render OK on a smaller window?

thisisrohan: This is just to make sure that the label 'velocity vectors' is displayed completely. It renders fine on a smaller screen, ipywidgets just creates a scroll bar to traverse all the particle arrays.

pysph/tools/ipy_viewer.py
thisisrohan: I've updated this method in the new commit, it no longer uses this function, and yes, it was supposed to be `all_plots/frame..`.

pysph/tools/ipy_viewer.py
thisisrohan: Corrected this in the new commit.

prabhuramachandran:
Thanks! Merging for now.

PR #176, opened 2019-02-26
Fix delta sph equations

abhinavmuta:


psnbaba:
Formulation seems corrects, should add these references
for,

    fac = 2 * (self.dim + 2) * self.nu * self.rho0 * piij * Vj / rhoi

add reference for the term ` 2 * (self.dim + 2) * self.nu   `

in `class ContinuityEquationDeltaSPH(Equation):` add comment that for $L_a$ is computed in the prestep and not missed in the equation.

PR #182, opened 2019-03-06
Fixed issues with legends, added capability to view simulation time and a visibility checkbox.

thisisrohan:
Fixed multiple issues with the legends (wrong colors, random values, ....) and minor issues with the 3DViewer. Added the capability to view the simulation time step, as well a checkbox to toggle visibility of respective ParticleArrays.


thisisrohan:
Also note that I've set the matplotlib backend to the ipympl backend (this eliminates the need for the user to manually do this using the magic command `%matplotlib ipympl`). However this needs to be done before we import `matplotlib.pyplot` and hence this violates the PEP 8 style guide.

pysph/tools/ipy_viewer.py
```diff
@@ -237,7 +251,9 @@ class Viewer2DWidgets(object):
 
     def __init__(self, file, file_count):
 
-        self.temp_data = load(file)['arrays']
+        self.temp_data = load(file)
```
prabhuramachandran: I don't like the use of "file" as it overwrites the builtin `file` type maybe use `filename`?

pysph/tools/ipy_viewer.py
```diff
@@ -343,6 +371,19 @@ def _create_widgets(self):
...
         widgets.delay_box.observe(self._delay_box_handler, 'value')
         widgets.save_all_plots.observe(self._save_all_plots_handler, 'value')
+        widgets.show_solver_time.observe(
+            self._show_solver_time_handler,
+            'value'
+        )
+
+        # PLEASE NOTE:
+        # All widget handlers take in 'change' as an argument. This is usually
+        # a dictionary conatining information about the widget and the change
```
prabhuramachandran: "containing"

pysph/tools/ipy_viewer.py
```diff
@@ -410,16 +456,24 @@ def interactive_plot(self):
...
-                colormap = getattr(mpl.cm, cmap)
-                sct = self._scatters[array_name]
-                cmax = np.max(c)
-                if cmax != 0:
-                    sct.set_facecolors(colormap(c*1.0/cmax))
+                colormap = getattr(
+                    plt.cm,
+                    pa_widgets.scalar_cmap.value
+                )
+                arr = self._normalized_for_colormaps(c)
```
prabhuramachandran: The comments below are a sign that a refactor is in order, instead can't you use this:
```
min_c, max_c, abs_max_c, c_norm = self._normalized_for_colormaps(c)
```
This eliminates the need for the comments and is easier to read.  Later on also `arr[2]` doesn't convey much but the variable name will make the code easier to read.

pysph/tools/ipy_viewer.py
```diff
@@ -486,83 +544,109 @@ def _frame_handler(self, change):
...
-                cmap = pa_widgets.scalar_cmap.value
-                colormap = getattr(mpl.cm, cmap)
-                cmax = np.max(c)
-                if cmax != 0:
-                    sct.set_facecolors(colormap(c*1.0/cmax))
+                colormap = getattr(
+                    plt.cm,
+                    pa_widgets.scalar_cmap.value
+                )
+                arr = self._normalized_for_colormaps(c)
```
prabhuramachandran: Same as above.

pysph/tools/ipy_viewer.py
```diff
@@ -575,33 +659,46 @@ def _adjust_axes(self):
...
+            c = getattr(
+                    temp_data[array_name],
+                    pa_widgets.scalar.value
+            )
+            colormap = getattr(
+                plt.cm,
+                pa_widgets.scalar_cmap.value
+            )
+            sct = self._scatters[array_name]
+            arr = self._normalized_for_colormaps(c)
```
prabhuramachandran: Same as above.

pysph/tools/ipy_viewer.py
```diff
@@ -634,29 +732,48 @@ def _legend_handler(self, change):
...
+
+                    arr = self._normalized_for_colormaps(c)
+                    minm = arr[0]
+                    maxm = arr[1]
+                    absmaxm = arr[2]
+
+                    # Do not change the order of the if-else #
+                    # block below. It is important that the  #
+                    # checks are performed in this order,    #
+                    # otherwise the 'boundaries' array will  #
```
prabhuramachandran: Not sure if this comment is needed as order does often matter in an if/elif block.

pysph/tools/ipy_viewer.py
```diff
@@ -634,29 +732,48 @@ def _legend_handler(self, change):
...
+        if len(self._cbars.keys()) == 0:
+            self._scatter_ax.set_position(
+                [0, 0, 1, 1]
+            )
+        if change is not None:
+            self.figure.show()
 
     def _save_figure_handler(self, change):
 
+        flag = True
```
prabhuramachandran: `flag` is not a good variable name.

pysph/tools/ipy_viewer.py
```diff
@@ -727,6 +854,84 @@ def _save_all_plots_handler(self, change):
...
+            )
+            c = getattr(
+                temp_data[array_name],
+                pa_widgets.scalar.value
+            )
+            colormap = getattr(
+                plt.cm,
+                pa_widgets.scalar_cmap.value
+            )
+            arr = self._normalized_for_colormaps(c)
```
prabhuramachandran: Cleanup.

pysph/tools/ipy_viewer.py
```diff
@@ -920,210 +1149,248 @@ def _create_widgets(self):
...
-                    self.cbars[array_name].set_label(
-                                    array_name+" : "+pa_widgets.scalar.value
-                            )
-            display(self.pltfigure)
+                        arr = self._normalized_for_colormaps(c)
+                        minm = arr[0]
+                        maxm = arr[1]
+                        absmaxm = arr[2]
+
+                        # Do not change the order of the if-else #
```
prabhuramachandran: Seems like a lot of copy-paste code is here, don't repeat code, try to write the code so you write things only once and reuse things as much as possible.  You can do this with a subclass for example.

pysph/tools/ipy_viewer.py
```diff
@@ -1189,8 +1459,76 @@ def _save_all_plots_handler(self, change):
...
-                pa_widgets.vector_size.disabled = True
-                pa_widgets.scalar_size.disabled = True
+                pa_widgets.scalar.disabled = False
+                pa_widgets.scalar_cmap.disabled = False
+                pa_widgets.velocity_vectors.disabled = False
+                pa_widgets.vector_size.disabled = False
+                pa_widgets.scalar_size.disabled = False
+                pa_widgets.is_visible.disabled = False
+
+    def _normalized_for_colormaps(self, data):
```
prabhuramachandran: This is the same function as before, Don't Repeat Yourself (DRY).  Please refactor this to be reused.

thisisrohan: I've added this function to the base class `Viewer`, from which both the `2DViewer` and the `3DViewer` classes inherit. This takes care of this issue.

pysph/tools/ipy_viewer.py
thisisrohan: Changed it to `file_was_saved`, improves readability as well.

pysph/tools/ipy_viewer.py
thisisrohan: Agreed, this does seem unnecessary. I've removed this comment in the latest commit.

pysph/tools/ipy_viewer.py
thisisrohan: Agreed, this does improve readability a lot. I've changed this as per your suggestion in the latest commit.

pysph/tools/ipy_viewer.py
thisisrohan: Corrected.

pysph/tools/ipy_viewer.py
thisisrohan: Changed to `file_name`.

pysph/tools/ipy_viewer.py
thisisrohan: All of this has been changed in the latest commit.

pysph/tools/ipy_viewer.py
thisisrohan: I've implemented most of your suggestions. However in certain functions like the `_legend_handler`, even though both the `3DViewer` and the `2DViewer` share some code, there are kinks (like changing the axes positions before specific events/to specific positions) that kind of make it tough to write reusable code.

prabhuramachandran: It may be difficult but you should try very hard to see if they can be written once, for example if it is just an option you can generalize the function to handle both cases.

prabhuramachandran: Of course, not at the expense of things being readable and sane.

prabhuramachandran:
I will merge this for now.  I think some kind of refactoring may be useful at a later point to maximize reuse.

PR #183, opened 2019-03-07
added set wall velocity before CE solids

psnbaba:


pysph/sph/wc/gtvf.py
```diff
@@ -590,7 +590,12 @@ def get_equations(self):
         )
         all = self.fluids + self.solids
 
-        eq1, stage1 = [], []
+        eq0, stage1 = [], []
+        for solid in self.solids:
+            eq0.append(SetWallVelocity(dest=solid, sources=self.fluids))
+        stage1.append(Group(equations=eq0, real=False))
```
prabhuramachandran: This should be done only if `eq0` is not an empty list, else you will have an empty group in the code!

psnbaba: Done.

PR #184, opened 2019-03-18
Basic implementation of Rigid body dynamics using rotation matrices.

dineshadepu:
1. Added a stack of cylinders example.
2. Used only rotation matrices.

### DIFF
--- a/pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
+++ b/pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
@@ -2,6 +2,8 @@
 Simulation of solid-fluid mixture flow using moving particle methods
 Shuai Zhang
 
+link: https://www.sciencedirect.com/science/article/pii/S0021999108006499
+
 Time: 7 minutes
 """
 from __future__ import print_function
@@ -174,7 +176,10 @@ class ZhangStackOfCylinders(Application):
         return x, y, body_id
 
     def geometry(self):
-        import matplotlib.pyplot as plt
+        import matplotlib
+        matplotlib.use('Agg')
+        from matplotlib import pyplot as plt
+
         # please run this function to know how
         # geometry looks like
         xc, yc, body_id = self.create_cylinders_stack()
@@ -198,6 +203,13 @@ class ZhangStackOfCylinders(Application):
                     pa.y += 14 * 1e-2
 
     def post_process(self):
+        """This function will run once per time step after the time step is
+        executed. For some time (self.wall_time), we will keep the wall near
+        the cylinders such that they settle down to equilibrium and replicate
+        the experiment.
+
+        By running the example it becomes much clear.
+        """
         if len(self.output_files) == 0:
             return
 
--- a/pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
+++ b/pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
@@ -1,0 +1,239 @@
+"""
+Simulation of solid-fluid mixture flow using moving particle methods
+Shuai Zhang
+
+Time: 7 minutes
+"""
+from __future__ import print_function
+import numpy as np
+
+# PySPH base and carray imports
+from pysph.base.kernels import CubicSpline
+
+from pysph.solver.solver import Solver
+from pysph.sph.integrator import EPECIntegrator
+
+from pysph.sph.equation import Group
+from pysph.solver.application import Application
+from pysph.sph.rigid_body import (BodyForce)
+from pysph.sph.rb_rotation_matrices import (
+    get_particle_array_rigid_body, SumUpExternalForces, RK2StepRigidBody,
+    RigidBodyCollision)
+from pysph.tools.geometry import (get_2d_circle, get_2d_tank)
+
+
+class ZhangStackOfCylinders(Application):
+    def initialize(self):
+        self.dam_length = 26 * 1e-2
+        self.dam_height = 26 * 1e-2
+        self.dam_spacing = 1e-3
+        self.dam_layers = 2
+        self.dam_rho = 2000.
+
+        self.cylinder_radius = 1. / 2. * 1e-2
+        self.cylinder_diameter = 1. * 1e-2
+        self.cylinder_spacing = 1e-3
+        self.cylinder_rho = 2000.
+
+        self.wall_height = 20 * 1e-2
+        self.wall_spacing = 1e-3
+        self.wall_layers = 2
+        self.wall_time = 0.3
+        self.wall_rho = 2000.
+
+        # simulation properties
+        self.hdx = 1.2
+        self.alpha = 0.1
+
+        # solver data
+        self.tf = 0.6 + self.wall_time
+        self.dt = 5e-5
+
+    def create_particles(self):
+        # get bodyid for each cylinder
+        xc, yc, body_id = self.create_cylinders_stack()
+        m = self.cylinder_rho * self.cylinder_spacing**2
+        h = self.hdx * self.cylinder_radius
+        rad_s = self.cylinder_spacing / 2.
+        V = self.cylinder_spacing**2
+        cylinders = get_particle_array_rigid_body(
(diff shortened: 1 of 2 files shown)


dineshadepu:
This is a simple implementation. There are many areas of improvement and will be done in future. 
1. Use quaternions
2. Use principal axis of inertia to reduce the MOI to diagonal form
3. Solving the dynamics in body frame
4.DEM force will fix the physics.

dineshadepu:
@abhinavmuta  and @psnbaba leave comments

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
```diff
@@ -0,0 +1,239 @@
...
+
+        x_six_1, y_six_1 = np.concatenate(
+            (xc1, xc2, xc3, xc4, xc5, xc6)), np.concatenate((yc1, yc2, yc3,
+                                                             yc4, yc5, yc6))
+        x_six_2, y_six_2 = x_six_1, y_six_1 + 2. * self.cylinder_diameter
+        x_six_3, y_six_3 = x_six_2, y_six_2 + 2. * self.cylinder_diameter
+
+        x_six, y_six = np.concatenate(
+            (x_six_1, x_six_2, x_six_3)), np.concatenate((y_six_1, y_six_2,
+                                                          y_six_3))
```
psnbaba: Have you though of doing this?
    
    r = 0.5
    px = 0.5 + 1.0 # pitch along x
    py = 2 * 2 * r # pitch along y
    x = []
    y = []
    for i in range(6):
        for j in range(5):
                c0, c1 = px * i, py * j
                xc1, yc1 = get_2d_circle(dx, r, [c0, c1])
                x = np.concatenate(x, xc)
                y = np.concatenate(y, yc)

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
```diff
@@ -0,0 +1,239 @@
+"""
+Simulation of solid-fluid mixture flow using moving particle methods
+Shuai Zhang
```
psnbaba: Can you please add full paper name or link for the test case.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,539 @@
...
+        r = pa.body_limits[2 * i + 1]
+
+        j = 3 * i
+
+        pa.dx0[l:r] = pa.x[l:r] - pa.xcm[j]
+        pa.dy0[l:r] = pa.y[l:r] - pa.xcm[j + 1]
+        pa.dz0[l:r] = pa.z[l:r] - pa.xcm[j + 2]
+
+
+class RigidBodyCollision(Equation):
```
psnbaba: please add reference

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
```diff
@@ -0,0 +1,239 @@
...
+        plt.show()
+
+    def post_step(self, solver):
+        t = solver.t
+        dt = solver.dt
+        T = self.wall_time
+        if (T - dt / 2) < t < (T + dt / 2):
+            for pa in self.particles:
+                if pa.name == 'wall':
+                    pa.y += 14 * 1e-2
```
psnbaba: Why this is done? Can you please add a reference to this too.

dineshadepu:
This can be merged

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
```diff
@@ -0,0 +1,251 @@
+"""
```
prabhuramachandran: The first line should be short and should include the run time.  See any of the examples for examples.

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
prabhuramachandran: I agree, why haven't you done this?  Much easier to read and less messy.

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
prabhuramachandran: Perhaps a comment here to explain the intent may be good?

pysph/examples/rb_rotation_matrices/stack_of_cylinders.py
```diff
@@ -0,0 +1,239 @@
...
+                for i in range(array.n_body[0]):
+                    cm_x += array.xcm[3 * i] * array.total_mass[i]
+                    cm_y += array.xcm[3 * i + 1] * array.total_mass[i]
+                cm_x = cm_x / np.sum(array.total_mass)
+                cm_y = cm_y / np.sum(array.total_mass)
+
+                system_x.append(cm_x / self.dam_length)
+                system_y.append(cm_y / self.dam_length)
+
+        import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
prabhuramachandran: Why is the above not done?  The `geometry` function has this but is never called.  The problem with doing the above is that it is possible that the example is run via ssh and then this will crash.  For example, ssh to a machine and try importing matplotlib's pyplot and it will crash because the backend may be set to Qt.  By doing the above, you ensure that the post processing will always work when mpl is available regardless of the system etc.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,542 @@
...
+    pa.add_constant('n_body', np.array([n_body], dtype=int))
+
+    # set the index limits of the body
+    limits = get_limits_of_rigid_bodies(pa)
+    pa.add_constant('body_limits', limits)
+
+    # no of bodies array for compyle parallel computing
+    pa.add_constant('bodies', np.arange(0, pa.n_body[0], 1))
+
+    # total mass of the body
```
prabhuramachandran: This comment is not useful, can be removed.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,542 @@
...
+    set_total_mass(pa)
+
+    # center of mass of the body
+    set_center_of_mass(pa)
```
prabhuramachandran: Superfluous comment.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,542 @@
...
+    """This method assumes the center of mass is already computed."""
+    for i in range(pa.n_body[0]):
+        xcm_i = pa.xcm[3 * i:3 * i + 3]
+        # left limit of body i
+        l = pa.body_limits[2 * i]
+        # right limit of body i
+        r = pa.body_limits[2 * i + 1]
+
+        I = np.zeros(9)
+        for j in range(l, r):
```
prabhuramachandran: This loop can also be completely vectorized with numpy.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,542 @@
...
+    I_all = np.array([])
+    for i in range(pa.n_body[0]):
+        xcm_i = pa.xcm[3 * i:3 * i + 3]
+        # left limit of body i
+        l = pa.body_limits[2 * i]
+        # right limit of body i
+        r = pa.body_limits[2 * i + 1]
+
+        I = np.zeros(9)
+        for j in range(l, r):
```
prabhuramachandran: Again this loop can be vectorized.

pysph/sph/rb_rotation_matrices.py
```diff
@@ -0,0 +1,542 @@
...
+
+                # save the current angular momentum
+                dst.ang_mom_0[j] = dst.ang_mom[j]
+
+            # save the current orientation
+            # print(dst.orientation_R)
+            for j in range(i9, i9 + 9):
+                dst.orientation_R_0[j] = dst.orientation_R[j]
+
+    def initialize(self):
```
prabhuramachandran: Instead of doing this, you can simply avoid the need for initialize by writing your own integrator.  This empty initialize is not ideal.

PR #191, opened 2019-03-22
added check for using arbitrary name in argument

psnbaba:
The viewer currently doesn't give any warning for the required filename format. Same is added in the PR.


pysph/tools/mayavi_viewer.py
```diff
@@ -1154,6 +1154,13 @@ def main(args=None):
                 scripts.append(arg)
                 continue
             elif arg.endswith(output_formats):
+                try:
+                    _sort_key(arg)
+                except ValueError:
+                    print("Error: file name is not supported")
+                    print("filename format accepted is *_number.npz"
+                          " or *_number.hdf5")
+                    exit(-1)
```
prabhuramachandran: Please change this to `sys.exit(1)`

PR #194, opened 2019-03-24
 Add update_minmax_cl method to calculate min and max in single scan

adityapb:


pysph/base/gpu_domain_manager.py
```diff
@@ -71,7 +71,7 @@ def _compute_cell_size_for_binning(self):
 
         for pa_wrapper in self.pa_wrappers:
             h = pa_wrapper.pa.gpu.get_device_array('h')
-            h.update_min_max()
+            pa_wrapper.pa.gpu.update_minmax_cl(['h'])
```
prabhuramachandran: It would be great if this and the `x, y, z` could be merged into one reduction.  `update_domain()` is invariably called before nnps update.  Thoughts?

prabhuramachandran:
Thanks!

PR #195, opened 2019-03-24
added gradient interpolation

psnbaba:
The interpolator doesn't give gradients in the current version.

### DIFF
--- a/pysph/sph/wc/linalg.py
+++ b/pysph/sph/wc/linalg.py
@@ -62,26 +62,31 @@ def mat_vec_mult(a=[1.0, 0.0], b=[1.0, 0.0], n=3, result=[0.0, 0.0]):
         result[i] = s
 
 
-def augmented_matrix(A=[0.0, 0.0], b=[0.0, 0.0], n=3, na=1, result=[0.0, 0.0]):
+def augmented_matrix(A=[0.0, 0.0], b=[0.0, 0.0], n=3, na=1, nmax=3,
+                     result=[0.0, 0.0]):
     """Create augmented matrix.
 
-    Given flattened matrix, `A`, and flattened columns `b` with `n` rols and
-    `na` additional columns, put these in result. Result must be already
-    allocated and be flattened.
+    Given flattened matrix, `A` of max rows/columns `nmax`, and flattened
+    columns `b` with `n` rows of interest and `na` additional columns, put
+    these in `result`. Result must be already allocated and be flattened.
+    The `result` will contain `(n + na)*n` first entries as the
+    augmented_matrix.
+
 
     Parameters
     ----------
     A: list: given matrix.
     b: list: additional columns to be augmented.
-    n: int : number of rows/columns in `A`.
+    n: int : number of rows/columns to use from `A`.
     na: int: number of added columns in `b`.
-    result: list: must have size of (n + na)*n.
+    nmax: int: the maximum dimension 'A'
+    result: list: must have size of (nmax + na)*n.
     """
     i, j, nt = declare('int', 3)
     nt = n + na
     for i in range(n):
         for j in range(n):
-            result[nt*i + j] = A[n * i + j]
+            result[nt*i + j] = A[nmax * i + j]
         for j in range(na):
             result[nt*i + n + j] = b[na*i + j]
 
@@ -97,7 +102,7 @@ def gj_solve(m=[1., 0.], n=3, nb=1, result=[0.0, 0.0]):
     ----------
 
     m : list: a flattened list representing the augmented matrix [A|b].
-    n : int: number of columns/rows of A.
+    n : int: number of columns/rows used from A in augmented_matrix.
     nb: int: number of columns added to A.
     result: list: with size n*nb
 
--- a/pysph/tools/interpolator.py
+++ b/pysph/tools/interpolator.py
@@ -8,9 +8,12 @@ import numpy as np
 from pysph.base.utils import get_particle_array
 from pysph.base.kernels import Gaussian
 from pysph.base.nnps import LinkedListNNPS as NNPS
-from pysph.sph.equation import Equation
+from pysph.sph.equation import Equation, Group
 from pysph.sph.acceleration_eval import AccelerationEval
 from pysph.sph.sph_compiler import SPHCompiler
+from compyle.api import declare
+from pysph.sph.wc.linalg import gj_solve, augmented_matrix
+from pysph.sph.basic_equations import SummationDensity
 
 
 class InterpolateFunction(Equation):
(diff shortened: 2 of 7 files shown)


pysph/sph/wc/linalg.py
```diff
@@ -62,7 +62,8 @@ def mat_vec_mult(a=[1.0, 0.0], b=[1.0, 0.0], n=3, result=[0.0, 0.0]):
         result[i] = s
 
 
-def augmented_matrix(A=[0.0, 0.0], b=[0.0, 0.0], n=3, na=1, result=[0.0, 0.0]):
+def augmented_matrix(A=[0.0, 0.0], b=[0.0, 0.0], n=3, na=1, nmax=3,
```
prabhuramachandran: What is nmax, it should be described in the docstring and clearly explained.  Why and where is this needed?

pysph/tools/interpolator.py
```diff
@@ -27,12 +30,134 @@ def post_loop(self, d_idx, d_prop, d_number_density):
             d_prop[d_idx] /= d_number_density[d_idx]
 
 
-class InterpolateSPH(Equation):
-    def initialize(self, d_idx, d_prop):
-        d_prop[d_idx] = 0.0
+class UpdateD(Equation):
```
prabhuramachandran: A nicer name is necessary, I cannot figure out what the intent of this name is.

pysph/tools/interpolator.py
```diff
@@ -285,10 +420,11 @@ def _create_particle_array(self, x, y, z):
             x=xr, y=yr, z=zr, h=h,
             number_density=np.zeros_like(xr),
             prop=prop,
-            grad_x=np.zeros_like(xr),
-            grad_y=np.zeros_like(xr),
-            grad_z=np.zeros_like(xr)
         )
+        if not self.use_shepard:
+            pa.add_property('A', stride=16)
```
prabhuramachandran: These names are rather common, for example if the input array had a property called 'A' this would not work.  Can you use a more unique name and something a bit more meaningful.  Same for B and `propxyz`?

pysph/tools/interpolator.py
prabhuramachandran: A docstring on how this formula was arrived at or a link to a suitable reference is important.

pysph/sph/wc/linalg.py
psnbaba: done

pysph/tools/interpolator.py
psnbaba: added refs and changed var names

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
```diff
@@ -27,12 +30,113 @@ def post_loop(self, d_idx, d_prop, d_number_density):
...
+        for i in range(4):
+            for j in range(4):
+                d_mommat[16*d_idx + j+4*i] = 0.0
+
+    def loop(self, d_idx, d_h, s_h, s_x, s_y, s_z, d_x, d_y, d_z, s_rho,
+             s_m, WIJ, XIJ, DWIJ, d_mommat):
+        # Vj = s_m[s_idx] / s_rho[s_idx]
+        Vj = 1.0
+
+        d_mommat[16*d_idx+0] += WIJ * Vj
```
prabhuramachandran: You could pre-compute 16*d_idx into a variable and save 16 multiplications.

pysph/tools/interpolator.py
```diff
@@ -27,12 +30,113 @@ def post_loop(self, d_idx, d_prop, d_number_density):
...
+        self.dim = dim
+
+        super(SPHFirstOrderApproximationPreStep, self).__init__(dest, sources)
+
+    def initialize(self, d_idx, d_mommat):
+        i, j = declare('int', 2)
+
+        for i in range(4):
+            for j in range(4):
+                d_mommat[16*d_idx + j+4*i] = 0.0
```
prabhuramachandran: Instead of `mommat`, can't it just be `moments`?

pysph/tools/interpolator.py
```diff
@@ -27,12 +30,113 @@ def post_loop(self, d_idx, d_prop, d_number_density):
...
+
+    def loop(self, d_idx, d_h, s_h, s_x, s_y, s_z, d_x, d_y, d_z, s_rho,
+             s_m, WIJ, XIJ, DWIJ, d_mommat):
+        # Vj = s_m[s_idx] / s_rho[s_idx]
```
prabhuramachandran: Remove the commented code please.

pysph/tools/interpolator.py
```diff
@@ -209,19 +313,30 @@ def interpolate(self, prop, gradient=False):
...
         """
+        assert isinstance(comp, int), 'Error: only interger value is allowed'
         for array in self.particle_arrays:
             data = array.get(prop, only_real_particles=False)
             array.get('temp_prop', only_real_particles=False)[:] = data
 
         self.func_eval.compute(0.0, 0.1)  # These are junk arguments.
-        result = self.pa.prop.copy()
+        if comp and self.use_shepard:
+            print("Error: set use_shepard=False to evaluate gradients")
```
prabhuramachandran: Raise an exception here, you could for example do `raise RuntimeError('Set use_shepard=False ...')`, otherwise someone may choose to ignore the warning and accept the result as correct.

pysph/tools/interpolator.py
```diff
@@ -209,19 +313,30 @@ def interpolate(self, prop, gradient=False):
...
+            result = np.zeros_like(self.pa.x)
+        elif self.use_shepard:
```
prabhuramachandran: I think it makes sense to change the `use_shepard` option to an `order` or `method` option which has three possibilities ('sph', 'shepard', 'order1') and you can add an assert statement to ensure this in the `__init__`, you can keep `'shepard'` as the default.

pysph/tools/interpolator.py
```diff
@@ -279,27 +394,45 @@ def _create_particle_array(self, x, y, z):
...
         names = [x.name for x in self.particle_arrays]
         if self.equations is None:
             if self.use_shepard:
-                equations = [InterpolateFunction(dest='interpolate',
-                                                 sources=names)]
+                equations = [
+                    InterpolateFunction(dest='interpolate',
+                                        sources=names)
+                ]
             else:
```
prabhuramachandran: So here the original SPH interpolation is missing, this can be retained.

psnbaba: done

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
psnbaba: done

pysph/tools/interpolator.py
```diff
@@ -209,19 +325,29 @@ def interpolate(self, prop, gradient=False):
...
         """
+        assert isinstance(comp, int), 'Error: only interger value is allowed'
         for array in self.particle_arrays:
             data = array.get(prop, only_real_particles=False)
             array.get('temp_prop', only_real_particles=False)[:] = data
 
         self.func_eval.compute(0.0, 0.1)  # These are junk arguments.
-        result = self.pa.prop.copy()
+        if comp and (self.method in ['sph', 'shepard']):
+            raise RuntimeError("Error: used 'order1' method to evaluate"
```
prabhuramachandran: used -> use

Issue #155, opened 2018-12-18
add_particle adds particle with inconsistent array length when property have more than one stride.


pawansnegi:
The following code explains what is happening.

```
from pysph.base.utils import get_particle_array_wcsph as gpaw
import numpy as np

x = np.arange(0,1,0.1)
y = np.copy(x)
fluid = gpaw(x=x, y=y)
fluid.add_property('newprop', stride=3)
pa_add = {}
for prop, array in fluid.get_property_arrays().items():
    pa_add[prop] = np.array(array[0:5])
fluid.add_particles(**pa_add)
fluid.get_property_arrays()

print(len(fluid.x), len(fluid.newprop), fluid.stride)
```

here the add_particles should show error that the property array lengths are inconsistent instead of adding them.

prabhuramachandran:
Closing since this seems addressed to the extent possible.

PR #196, opened 2019-03-26
added fix for #155

psnbaba:


pysph/base/particle_array.pyx
```diff
@@ -569,6 +569,13 @@ cdef class ParticleArray:
         prop = ''
         for prop in particle_props:
             self._check_property(prop)
+            if prop in self.stride.keys():
+                stride = self.stride[prop]
+                prop_len = len(particle_props[prop])
+                base_len = len(particle_props['x'])
```
prabhuramachandran: You can put this out of the loop and find the number of real particles as self.get_number_of_particles(), do you need the ghost particles also?

pysph/base/particle_array.pyx
```diff
@@ -569,6 +569,13 @@ cdef class ParticleArray:
...
+                if not prop_len == base_len * stride:
+                    raise AttributeError, 'property %s length should be %d'%(prop,
```
prabhuramachandran: Use `raise AttributeError(...)`

pysph/base/particle_array.pyx
psnbaba: I have put this inside loop as only the `prop` in `particle_props` need to be checked for the consistency. Poperties which are not present in `particle_props` will be initialized consistently with the stride value later in the function. No, I don't need ghost particles.

`self.get_number_of_particles()` will give me length of the current particle array. However, I want the length of `x` in the new particle_array which is passed.

prabhuramachandran: You also do not need to do `if prop in self.stride.keys`, just `prop in self.stride` will do.

prabhuramachandran: I mean you can get the base_len outside the loop.  Also what if the particle_props do not have 'x'?  What then?

psnbaba: the new particles to be added should at least contain x, y and z. isn't it?
if not we should add that check too.

psnbaba:
> There are still issues with this that I am not comfortable with, also the gpu cases need to be addressed.

actually I was getting garbage value due to this bug and took a lot of time to find the bug as I was new to pysph. So this need to be fixed. I try to write test cases. Also if @adityapb have time, I can use his help.

adityapb:
> > There are still issues with this that I am not comfortable with, also the gpu cases need to be addressed.
> 
> actually, I was getting garbage value due to this bug and took a lot of time to find the bug as I was new to pysph. So this need to be fixed. I try to write test cases. Also if @adityapb have time, I can use his help for gpu cases.

@psnbaba I won't be on campus this week. We can meet on Monday and discuss this

pysph/base/particle_array.pyx
```diff
@@ -569,6 +569,14 @@ cdef class ParticleArray:
         prop = ''
         for prop in particle_props:
             self._check_property(prop)
+            if prop in self.stride.keys():
+                stride = self.stride[prop]
+                prop_len = len(particle_props[prop])
+                base_len = self.get_number_of_particles()
+                print(stride, prop_len, base_len)
```
prabhuramachandran: Remove the print statement.

pysph/base/particle_array.pyx
```diff
@@ -569,6 +569,14 @@ cdef class ParticleArray:
...
+                if not prop_len == base_len * stride:
```
prabhuramachandran: Write this as:
```
msg = 'Property %s length should be %d' % (prop, base_len * stride)
assert prop_len == base_len * stride, msg
```

prabhuramachandran:
Also, the tests are failing on the GPU as in that case this check is not performed and an assertion error is not raised.

pysph/base/particle_array.pyx
psnbaba: Done

pysph/base/particle_array.pyx
psnbaba: Done

PR #201, opened 2019-04-09
added inviscid wall

psnbaba:


pysph/sph/wc/edac.py
```diff
@@ -734,6 +746,16 @@ def _get_internal_flow_equations(self):
                                     gx=self.gx, gy=self.gy, gz=self.gz),
                 SetWallVelocity(dest=solid, sources=fluids_with_io),
             ])
+        for solid in self.inviscid_solids:
+            group1.extend([
+                NoSlipVelocityExtrapolation(
```
prabhuramachandran: Since you are using these together, why not do both in the same equation, in fact, this will give you an error as `d_wij` will be added to twice, once for each equation leading to incorrect solid wall velocities.

prabhuramachandran: We need to make sure this mistake is not happening already as we use d_wij already in a few places.

pysph/sph/wc/edac.py
```diff
@@ -809,6 +833,15 @@ def _get_external_flow_equations(self):
                     ClampWallPressure(dest=solid, sources=None)
                 )
 
+        for solid in self.inviscid_solids:
+            group1.extend([
+                NoSlipVelocityExtrapolation(
+                    dest=solid, sources=self.fluids),
+                VolumeSummation(dest=solid, sources=all),
+                SolidWallPressureBC(dest=solid, sources=self.fluids,
```
prabhuramachandran: Here also there is a problem as both the first and last equations update d_wij.

pysph/sph/wc/transport_velocity.py
```diff
@@ -734,3 +734,72 @@ def post_loop(self, d_idx, d_wij, d_p, d_rho):
 
         # update the density from the pressure Eq. (28)
         d_rho[d_idx] = self.rho0 * (d_p[d_idx]/self.p0 + self.b)
+
+
+class NoSlipVelocityExtrapolation(Equation):
+    '''No Slip boundary condition on the wall
```
prabhuramachandran: Is this really no-slip or is it a slip-wall?

prabhuramachandran: Doesn't it depend on whether viscosity is used between this wall and the fluid?

psnbaba: It is slip wall.

Yes, but we cannot just disbale viscosity. the wall velocity needs to be same as fluid so that not gradient develops.

pysph/sph/wc/edac.py
psnbaba: yes I will look into it.

PR #203, opened 2019-04-19
Added support for strided arrays in the notebook viewer.

thisisrohan:
Added a 'component' IntBox in the notebook viewer to allow users to view components of strided arrays.


prabhuramachandran:
LGTM, thanks.

PR #207, opened 2019-04-26
Gas dynamics fixes

ananyo-work:
- minor fixes in GSPHAcceleration
 - copy properties to periodic ghosts in GSPHScheme
 - add acoustic wave diffusion examples (1 and 2-D)
 - expose density iteration parameters in GasDScheme (MPM)
 - fix small bug with vsig calculation in MPMAcceleration

### DIFF
--- a/pysph/examples/gas_dynamics/accuracy_test_2d.py
+++ b/pysph/examples/gas_dynamics/accuracy_test_2d.py
@@ -1,0 +1,166 @@
+"""
+Two Dimensional constant pressure accuracy test
+particles should simply advect in a periodic domain
+"""
+
+from __future__ import print_function
+# NumPy and standard library imports
+import numpy
+
+from pysph.base.nnps import DomainManager
+from pysph.base.utils import get_particle_array as gpa
+from pysph.solver.application import Application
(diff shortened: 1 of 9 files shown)


pysph/examples/gas_dynamics/accuracy_test_2d.py
```diff
@@ -0,0 +1,165 @@
+"""
```
prabhuramachandran: Each example's docstring should follow a simple convention, see this for example:
https://github.com/pypr/pysph/blob/master/pysph/examples/dam_break_2d.py

The first line basically should be a short description with an indication of the time required to run the example on a single core.

pysph/examples/gas_dynamics/acoustic_wave.py
```diff
@@ -0,0 +1,129 @@
+"""
```
prabhuramachandran: Same here and elsewhere below.

prabhuramachandran:
Thanks, looks good.  In a subsequent PR, please do update the docstrings to use the right math markup and use raw strings.

PR #209, opened 2019-04-30
Rigid body dynamics using Rotation matrices and quaternions.

dineshadepu:


dineshadepu:
Will implement a better version. Closing for now.

PR #212, opened 2019-05-16
Mirrorbc

ananyo-work:
- add mirrored bc to DomainManager (CPU only)
  - add examples woodward-collela blastwave (with exact) and 2d riemann problem
  - update docstring for examples


pysph/base/nnps_base.pyx
```diff
@@ -262,7 +265,8 @@ cdef class DomainManager:
         """
         self.backend = get_backend(backend)
         is_periodic = periodic_in_x or periodic_in_y or periodic_in_z
-        if self.backend is 'opencl' or self.backend is 'cuda':
+        is_mirror = mirror_in_x or mirror_in_y or mirror_in_z
+        if (self.backend is 'opencl' or self.backend is 'cuda') and not is_mirror:
             from pysph.base.gpu_domain_manager import GPUDomainManager
             domain_manager = GPUDomainManager
```
prabhuramachandran: You should print a warning that the GPU backend does not support mirror BCs.

pysph/examples/gas_dynamics/riemann_2d.py
```diff
@@ -0,0 +1,257 @@
+"""2-d Riemann problem (xx minutes)
```
prabhuramachandran: Fill in xx.

prabhuramachandran: The time estimate is for the default parameters with no additional command line args by the way.

PR #215, opened 2019-05-26
Preliminary CUDA support

prabhuramachandran:
This PR adds basic CUDA support. While this is functional, it is not tested as well as OpenCL.  It does not support the different NNPS algorithms that are available yet. It does include some fixes for OpenCL support.


adityapb:
Looks good to me

PR #225, opened 2019-06-20
CRKSPH scheme, MPM and mirrorbc fix

ananyo-work:
- CRKSPH initial working scheme
  - Fix mirrorbc update
  - Add ADKE/MPM ghost particle property update
  - Fix bug with MPM
  - Add hydrostatic box example
  - Add Kelvin-Helmholtz instability example
  - Fix Noh's cylindrical implosion example


pysph/sph/wc/crksph.py
```diff
@@ -625,11 +630,10 @@ def loop(self, d_idx, s_idx, d_m, s_m, d_rho, s_rho, d_p, s_p, d_cs, s_cs,
         Qi = rhoi * (-Cl*ci*mui + Cq*mui*mui)
         Qj = rhoj * (-Cl*cj*muj + Cq*muj*muj)
 
-        fac = -(0.5/mi) * Vi * Vj * (pi + pj + Qi + Qj)
-
-        d_au[d_idx] += fac * DWIJ[0]
-        d_av[d_idx] += fac * DWIJ[1]
-        d_aw[d_idx] += fac * DWIJ[2]
+        fac = -(1.0/mi) * Vi * Vj * (pi + pj + Qi + Qj)
+        d_au[d_idx] += fac * (DWIJ[0])  # - DWJ[0])
```
prabhuramachandran: Why is the anti-symmetric component commented out?  Is that a mistake?

pysph/sph/wc/crksph.py
```diff
@@ -674,14 +688,88 @@ def initialize(self, d_idx, d_ae):
...
+        muj = min(0, tmpmuj)
+
+        Qi = rhoi * (-Cl*ci*mui + Cq*mui*mui)
+        Qj = rhoj * (-Cl*cj*muj + Cq*muj*muj)
+
+        fac = -(1.0 / mi) * Vi * Vj * (pi + pj + Qi + Qj)
+
         gamma = self.gamma
-        d = declare('int')
+        # d = declare('int')
```
prabhuramachandran: Please remove irrelevant comments.

pysph/sph/wc/crksph.py
```diff
@@ -741,9 +872,47 @@ def get_particle_array_crksph(constants=None, **props):
...
+class CRKSPHIntegrator(Integrator):
+    def one_timestep(self, t, dt):
+        self.stage1()
+        self.do_post_stage(dt, 1)
+
+        self.compute_accelerations(0)
+
+        self.stage2()
+        # We update domain here alone as positions only change here.
+        # self.update_domain()
```
prabhuramachandran: Why is this commented?

pysph/sph/wc/crksph.py
```diff
@@ -781,20 +950,24 @@ def configure_solver(self, kernel=None, integrator_cls=None,
             Any additional keyword args are passed to the solver instance.
         """
         from pysph.base.kernels import QuinticSpline
-        from pysph.sph.integrator import PECIntegrator
-        from pysph.sph.integrator_step import WCSPHStep
+        # from pysph.sph.integrator import PECIntegrator
```
prabhuramachandran: If not used, please remove.

pysph/sph/scheme.py
```diff
@@ -1087,7 +1088,14 @@ def get_equations(self):
         g3 = []
         for solid in self.solids:
             g3.append(WallBoundary(solid, sources=self.fluids))
-            equations.append(Group(equations=g3))
+        equations.append(Group(equations=g3))
+
+        gh = []
+        for fluid in self.fluids:
```
prabhuramachandran: The ghost stuff should be added only if a new kwarg called `has_ghosts` is set to True.  What if someone is not using ghosts?

pysph/sph/wc/crksph.py
```diff
@@ -804,73 +977,163 @@ def configure_solver(self, kernel=None, integrator_cls=None,
...
+                StateEquation(dest=fluid, sources=None, gamma=self.gamma)
+            )
+            eq00.append(
+                SpeedOfSound(dest=fluid, sources=None, gamma=self.gamma)
+            )
+        equations_stage1.append(Group(equations=eq00))
+
+        gh = []
+        for fluid in self.fluids:
+            gh.append(
```
prabhuramachandran: Same here, if there is no ghost this should not be done, right?  Unless this is safe to use and we are guaranteed that orig_idx is correctly set?

pysph/sph/wc/crksph.py
ananyo-work: DWIJ = 0.5*(DWI - DWJ)

i was just testing earlier, will remove the comment

pysph/sph/wc/crksph.py
ananyo-work: because we only update velocity in stage2() for the energy equation, after the update is done, we use stage3() to update the position and energy, will clean this up

pysph/sph/scheme.py
ananyo-work: if someone is using the predefined scheme and has not overloaded 
    
    def setup_properties(self, particles, clean=True):
        ...

it will setup the `orig_idx`, but we can have a check for ghosts, will save some computation time checking if a particle in domain is a ghost

pysph/examples/gas_dynamics/acoustic_wave.py
```diff
@@ -36,20 +36,31 @@ def initialize(self):
...
         self.dim = 1
 
     def create_domain(self):
         return DomainManager(
             xmin=0, xmax=1, periodic_in_x=True
         )
 
+    def add_user_options(self, group):
+        group.add_argument(
+            "--nparticles", action="store", type=int, dest="nprt", default=256,
```
prabhuramachandran: How about `--np`` instead?

PR #229, opened 2019-07-13
added option to neglect command line arguments

psnbaba:
When an application is called inside an application, it takes the command line argument which is not required always. e.g

        from case_2d_bounded_airfoil import Airfoil2d
        airfoil = Airfoil2d(noparse=True)
        airfoil.scheme.solver.tf = 0.1
        airfoil.scheme.solver.pfreq = 1e14
        airfoil.run()
        xs, ys, xf, yf = airfoil.scheme.post_process(
            airfoil.particles, airfoil.boundary, 'internal', 'airfoil')
        print("Airfoil created...")
        fluid = self._create_fluid(xf, yf)
        solid = self._create_airfoil(xs, ys)

Here a airfoil is created using another application and used to creare partcile in current app. Now, previously it took arguments from the command line even if pfreq and tf was set here.

### DIFF
--- a/pysph/solver/application.py
+++ b/pysph/solver/application.py
@@ -173,7 +173,8 @@ class Application(object):
 
     """
 
-    def __init__(self, fname=None, output_dir=None, domain=None):
+    def __init__(self, fname=None, output_dir=None, domain=None,
+                 no_parse=False):
         """ Constructor
 
         Parameters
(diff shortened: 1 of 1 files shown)


psnbaba:
call app.run([]) to ignore command line args.

PR #233, opened 2019-08-26
Added 'cull', 'binder' and 'Viewer1D'.

thisisrohan:
Added the commands 'cull' and 'binder', to reduce a directory's size and make a mybinder.org compatible directory (respectively). Added 'Viewer1D' in the ipy_viewer, to enable viewing 1D simulations. Minor bug fixes.


pysph/tools/binder.py
```diff
@@ -0,0 +1,208 @@
...
+    '''
+    Finds the type of viewer to use in the jupyter notebook.
+    Parses the log file at the file path and searches for
+    'dim=d', where 'dD' is taken to be the viewer type.
+
+    example: if 'dim=2', then what is returned is '2D'
+    '''
+
+    log_file_path = os.path.abspath(path) + '/*.log'
+    regex = 'dim=(\d)'
```
prabhuramachandran: Use raw strings, i.e. `r'dim=(\d)'`, as discussed here: https://docs.python.org/3.7/library/re.html

pysph/tools/binder.py
```diff
@@ -0,0 +1,208 @@
...
+
+    nb = nbf.v4.new_notebook()
+
+    nb.cells = [
+        nbf.v4.new_code_cell(source=cell1_src),
+        nbf.v4.new_code_cell(source=cell2_src),
+        nbf.v4.new_code_cell(source=cell3_src)
+    ]
+
+    nbf.write(nb, path+'/'+sim_name+'.ipynb')
```
prabhuramachandran: Use `os.path.join` instead of doing the `'/'` additions.

pysph/tools/binder.py
```diff
@@ -0,0 +1,208 @@
...
+
+    sim_paths_list = find_sim_dirs(src_path)
+
+    for path in sim_paths_list:
+        sim_name = os.path.split(path)[1]
+        make_notebook(path, sim_name)
+        if len(sim_paths_list) == 1 and sim_paths_list[0] == src_path:
+            files = os.listdir(src_path)
+            files = [src_path+'/'+f for f in files]
+            mkdir(src_path+'/'+sim_name)
```
prabhuramachandran: Use `os.path.join` here and other places also.

pysph/tools/binder.py
```diff
@@ -0,0 +1,208 @@
...
+            could_not_copy = []
+            for f in files:
+                try:
+                    copy(f, src_path+'/'+sim_name)
+                except:
```
prabhuramachandran: Do not use catch-all except clauses as it will ignore any errors which can be dangerous.  Just catch the ones you need.

pysph/tools/binder.py
```diff
@@ -0,0 +1,208 @@
...
+            "(https://mybinder.org/v2/[hosting_service]/[repo_specifics]" +
+            "/[branch_name])\n" +
+            "\n" +
+            "[comment]: # (GitLab users: hosting_service = gl, " +
+            "repo_specifics = [user_name]%2F[repo_name])\n" +
+            "\n" +
+            "[comment]: # (GitHub users: hosting_service = gh, " +
+            "repo_specifics = [user_name]/[repo_name])\n" +
+            "\n" +
+            "[comment]: # (Gist users: hosting_service = gist, " +
```
prabhuramachandran: Instead of this comment, can you just add all three examples and have users delete/modify those lines:
```
![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gl/user_name/repo_name/master)
```
Or whatever is needed. For more cases, users can always read some documentation -- perhaps a link to the binder docs may be useful for them to quickly refer instead of having to google it.

pysph/tools/cull.py
```diff
@@ -0,0 +1,88 @@
...
+    src_path = os.path.abspath(src_path)
+    sim_paths_list = find_sim_dirs(src_path)
+
+    initial_size = find_dir_size(src_path)
+
+    for path in sim_paths_list:
+        files = get_files(path)
+        safe_files = [files[i] for i in range(0, len(files), c)]
+        for f in files:
+            if f in safe_files:
```
prabhuramachandran: This is a bit inefficient and may not matter but you are doing a lookup in a list each time, so you could make `safe_files` a `set` to make this more efficient or change the logic to not require a lookup.

pysph/tools/ipy_viewer.py
```diff
@@ -572,7 +581,684 @@ def _stride_and_component(self, data, pa_widgets):
...
+        '''
+        Helper function for making individual y-axes
+        for different particle arrays
+        '''
+        ax.set_frame_on(val)
+        ax.patch.set_visible(False)
+        for sp in ax.spines.values():
+            sp.set_visible(False)
+
+    def _frame_handler(self, change):
```
prabhuramachandran: There seems to be a lot of repeated code here and in the other Viewers.  Is there a way these can be abstracted out to minimize code repetition?

pysph/tools/binder.py
thisisrohan: Fixed.

pysph/tools/binder.py
thisisrohan: Fixed.

pysph/tools/binder.py
thisisrohan: Fixed.

pysph/tools/binder.py
thisisrohan: I've updated this block to where it captures all exceptions and prints them along with the files that could not be copied.

pysph/tools/binder.py
thisisrohan: I've changed dummy link to work for GitHub repos only, and added a statement saying that the user can go to mybinder.org for information on working with other hosting services.

pysph/tools/cull.py
thisisrohan: Fixed, changed the array such that it stores the paths of the files to be deleted.

pysph/tools/ipy_viewer.py
thisisrohan: I've refactored as much code as I could for the 2D and 3D Viewers, but the problem still remains that all three viewers use markedly different ways to display the plots. Not just this, the biggest issue I've found in refactoring code is that the way the legends are handled for the three viewers are completely different, and thus cannot be easily abstracted out.

pysph/tools/cull.py
```diff
@@ -0,0 +1,92 @@
...
+
+    src_path = os.path.abspath(src_path)
+    sim_paths_list = find_sim_dirs(src_path)
+
+    initial_size = find_dir_size(src_path)
+
+    for path in sim_paths_list:
+        files = get_files(path)
+        del_files = [
+            files[i] for i in np.setdiff1d(
```
prabhuramachandran: You don't need numpy here for this, the standard library `set` will do the trick just as well.
```
In [1]: set(range(5)) - set(range(0, 5, 2))
Out[1]: {1, 3}
```

pysph/tools/ipy_viewer.py
prabhuramachandran: OK.

pysph/tools/binder.py
prabhuramachandran: Can you use https://mybinder.readthedocs.io instead of just mybinder.org so it is easier to get to.

pysph/tools/cull.py
thisisrohan: Changed it in the latest commit.

pysph/tools/binder.py
thisisrohan: Changed it in the latest commit.

PR #241, opened 2019-09-09
added check particle array size

psnbaba:
Sometimes in inlet/outlet/fluid we may have zero particles. There without checking the particle size going for min of the array will gice error.

### DIFF
--- a/pysph/sph/integrator.py
+++ b/pysph/sph/integrator.py
@@ -86,7 +86,7 @@ class Integrator(object):
                 'dt_adapt' in pa.properties for pa in a_eval.particle_arrays
             )
         if self._has_dt_adapt:
-            dt_min = 1e20
+            dt_min = np.inf
             for pa in a_eval.particle_arrays:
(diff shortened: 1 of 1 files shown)


pysph/sph/integrator.py
```diff
@@ -90,11 +90,18 @@ def _get_explicit_dt_adapt(self):
             for pa in a_eval.particle_arrays:
                 if 'dt_adapt' in pa.properties:
                     if pa.gpu is not None:
-                        from compyle.array import minimum
-                        min_val = minimum(pa.gpu.dt_adapt)
+                        if pa.gpu.get_number_of_particles() > 0:
+                            from compyle.array import minimum
+                            min_val = minimum(pa.gpu.dt_adapt)
+                        else:
+                            min_val = 1e20
```
prabhuramachandran: I think this value and the `dt_min` value above should be set to `np.inf` instead of this arbitrary 1e20 number.

psnbaba: done

PR #242, opened 2019-09-11
added check to dump detailed output at the end

psnbaba:
credit @dineshadepu


psnbaba:
Have to think about it more closely. Closing for now.

PR #244, opened 2019-09-30
bug fix

psnbaba:
taking length from existing particle is wrong. We should check length from input particle array.
Also all particles to be added must have `h` thus we can safely take length of `h` for comparison.

### DIFF
--- a/pysph/base/particle_array.pyx
+++ b/pysph/base/particle_array.pyx
@@ -554,12 +554,6 @@ cdef class ParticleArray:
         prop = ''
         for prop in particle_props:
             self._check_property(prop)
-            if prop in self.stride.keys():
-                stride = self.stride[prop]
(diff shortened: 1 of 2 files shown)


pysph/base/particle_array.pyx
```diff
@@ -557,7 +557,7 @@ cdef class ParticleArray:
             if prop in self.stride.keys():
                 stride = self.stride[prop]
                 prop_len = len(particle_props[prop])
-                base_len = self.get_number_of_particles()
+                base_len = len(particle_props['h'])
```
prabhuramachandran: In general we cannot be sure that `h` exists as a prop in some cases, so perhaps pick any property with a unit stride and find its length.  Also, are we guaranteed that the particle props being passed are one dimensional i.e. `len(prop.shape) ==1`?

PR #245, opened 2019-10-24
Minor Defect Overwriting User Input

jeffreydaye:
When helping a coworker run the flow_past_cylinder_2d example it was noticed that changing the wind tunnel length or width resulted in no change to the simulation. Upon code inspection it was noted that the user input was overwritten by literal values.


prabhuramachandran:
Thanks for the fix!  Merging.

PR #253, opened 2020-04-17
Add extended uniform grid SFC and more NNPS changes

adityapb:
These are the algorithms used for the NNPS paper. Mainly 2 changes:
1. Added extended uniform grid based on SFC
2. Used arrays instead of maps for storing key to index maps. Makes it faster but memory inefficient for sparse distributions, just like linked list.

The extended uniform grid on SFC currently does not work for separate source and destination indices. I'll fix it in a later PR.


prabhuramachandran:
Hey @adityapb -- can you please resolve the conflicts so I can take a look? Thanks.

adityapb:
See #260

PR #269, opened 2020-05-21
Fix issue with merging props in interpolator.

abhinavmuta:
Added type and stride info while merging the props. Currently strides cannot be interpolated.


prabhuramachandran:
The merging of props seems all wrong, and this was written 6 years ago so I think a better solution is needed.

prabhuramachandran:
Thanks, yes way better.

PR #270, opened 2020-05-24
Optimization for stratified SFC NNPS

adityapb:
Calculate neighbor boxes for bins instead of particles
Makes stratified SFC about 1.5x faster than octree in case of variable h with max/min ratio of 4

### DIFF
--- a/pysph/base/stratified_sfc_nnps.pyx
+++ b/pysph/base/stratified_sfc_nnps.pyx
@@ -114,6 +114,7 @@ cdef class StratifiedSFCNNPS(NNPS):
         cdef int** current_key_to_idx
         cdef int** current_key_to_nbr_idx
         cdef int** current_key_to_nbr_length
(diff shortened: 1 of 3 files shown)


pysph/base/stratified_sfc_nnps.pyx
```diff
@@ -69,41 +72,74 @@ cdef class StratifiedSFCNNPS(NNPS):
         self.keys = <uint64_t**> malloc(narrays*sizeof(uint64_t*))
         self.key_to_idx = <int***> malloc(narrays*sizeof(int**))
         self.cell_sizes = <double**> malloc(narrays*sizeof(double*))
+        self.hmax = <double***> malloc(narrays*sizeof(double**))
```
prabhuramachandran, line 75: This doesn't seem to be free'd.  Can you please check and ensure there are no memory leaks?

PR #272, opened 2020-06-22
Sloshing Tank and Dam break Examples 

amalss18:
Examples added:
1) Sloshing under Horizontal Excitement based on Faltinsen et al., 2000
2) Sloshing due to pitch motion based on Akyildiz et al., 2006
3) Shallow water sloshing based on Bouscasse et al., 2013
4) Dam break against a tall structure based on an experiment by Yeh and Petroff (refer to Raad et al., 2005)


pysph/examples/db_3d_yeh.py
```diff
@@ -0,0 +1,165 @@
...
+        dx = self.options.dx
+        self.dx = dx
+        self.hdx = self.options.hdx
+        self.h0 = self.hdx * self.dx
+        self.geom = DamBreak3DGeometry(
+            container_height=0.4, container_width=0.61, container_length=1.6,
+            fluid_column_height=0.3, fluid_column_width=0.61, fluid_column_length=0.4,
+            obstacle_center_x=0.96, obstacle_center_y=0,
+            obstacle_length=0.12, obstacle_height=0.75, obstacle_width=0.12,
+            nboundary_layers=2, with_obstacle=True, dx=dx, hdx=hdx, rho0=ro
```
abhinavmuta, line 63: Use the global variables that you declared above for the parameters.

pysph/examples/db_3d_yeh.py
```diff
@@ -0,0 +1,165 @@
...
+            ['fluid'], ['boundary', 'obstacle'], dim=dim, rho0=ro, c0=c0,
+            h0=h0, hdx=hdx, gz=-9.81, alpha=alpha, beta=beta, gamma=gamma,
+            hg_correction=True, delta_sph=True, delta=0.1
+        )
+        return s
+
+    def configure_scheme(self):
+        s = self.scheme
+        hdx = self.hdx
+        kernel = Gaussian(dim=dim)
```
abhinavmuta, line 78: Why is Gaussian kernel the default? Could we not use cubic/quintic spline?

amalss18: δ-SPH model for simulating violent impact flows (Marrone et al., 2011) has implemented the same experiment using a δ-SPH Scheme and a Gaussian kernel. Shall I change it to Cubic Spline?

pysph/examples/st_bouscasse.py
```diff
@@ -0,0 +1,190 @@
...
+        super(HorizontalExcitation, self).__init__(dest, sources)
+
+    def initialize(self, d_idx, d_au, t, d_rho):
+        L = 1
+        h = 0.03 * L
+        A = 2.333 * h
+
+        k = pi/L
+        omega_r = sqrt(9.81*k*tanh(k*h))
+        omega = omega_r*1.231
```
abhinavmuta, line 64: Pass the variables to `__init__` as equation parameters, don't declare them here. Avoid using single letter names to variables.

pysph/examples/db_3d_yeh.py
abhinavmuta: Oh was not aware of it. Yes you can use the same kernel as the original paper or change it if you find Cubic spline better than Gaussian.

pysph/examples/sloshing_tank_horizontal.py
```diff
@@ -48,16 +47,22 @@
...
 
+length = 1.73
+h_tank = 1.15
+h_liquid = 0.6
+
 
 class HorizontalExcitation(Equation):
     def __init__(self, dest, sources):
+        self.amp = 0.032
+        self.T = 1.3
```
abhinavmuta, line 58: Pass these variables to the `__init__` method so it can be configured from the equation.

pysph/examples/st_examples/sloshing_tank_pitch.py
```diff
@@ -45,14 +44,21 @@
...
+length = 0.92
+width = 0.46
+height = 0.62
+n_layers = 3
+
 
 class PitchingMotion(Equation):
     def __init__(self, dest, sources):
+        self.theta_0 = 4 * pi/180
+        self.omega_r = 2
```
abhinavmuta, line 56: Same here, pass these to the `__init__` method.

prabhuramachandran:
@amalss18 -- I don't like the file organization, could you use better directory names?  I don't like db_examples, instead say dam_break?  Also, `st_examples` to `sloshing`?  The new directories should have an __init__.py as well.

PR #274, opened 2020-07-18
Fix failing travis builds.

prabhuramachandran:


prabhuramachandran:
This was due to a compyle bug that will be fixed shortly.

PR #276, opened 2020-07-24
Simulation of flexible fibers.

meyer-nils:
Hi,

thank you for developing PySPH. It is a nice SPH implementation and good starting point for implementing own ideas. We used PySPH in our research to model flexible fibers and finally published a paper [1] and the code [2].

I am wondering, wether you would be interested in incorporating this feature to the PySPH main repository? 

Cheers,
Nils

[1] https://doi.org/10.3390/jcs4020077
[2] https://github.com/nilsmeyerkit/pysph


prabhuramachandran:
@nilsmeyerkit -- thanks for this!  This takes a lot of effort and time so I am grateful that you have taken the time to contribute these changes!  I have been completely swamped with other things this week and it looks like it will be this way until end of this week at least. I think I will only be able to take a close look this weekend.  On a quick look my current issues are the following:

- changing nnps_base and the domain manager is not a good idea.  A subclass would be better as this is highly specialized and is only used in your example and not by anything else.  Perhaps there is a design issue where we need to make it easier to extend some of these things.  The periodicity handling is not ideal currently.
- I see a few indentation issues in the pyx file that I will mark when I do the full review.
- This is minor but I am not sure we should add more get_particle_array_xxx to the utils.  As PySPH grows some of these really need to be cleaned up and initially they were convenient but now as the repository grows there seem to be too many of these.

nilsmeyerkit:
@prabhuramachandran - thank you for your feedback. I can certainly make some improvements based on your full review. Though, it might take a while as this is just a side project of mine. 

Thanks
Nils

prabhuramachandran:
Hi @nilsmeyerkit -- I am terribly sorry for completely dropping the ball on this. As I go over the PR again now, I am not sure it makes sense to incorporate all of this directly into pysph because I am not sure I will be able to maintain and fix this feature in the future.  For all of our papers in https://gitlab.com/pypr, we only pull in changes into pysph if the functionality is general enough that it warrants inclusion in PySPH or if some basic feature is needed in pysph that is required.  Given this, do you think there is any generic functionality that is essential for your research that we should incorporate into pysph?  I will be happy to discuss that.  What do you think?  My apologies again for not getting back on this.

nilsmeyerkit:
Dear Prabhu, 
I understand that this is a rather niche topic. A general feature that might be interesting for multiple pysph users could be the incorporation of Lees-Edwards boundary conditions. I use them to apply shear to a completely periodic cube and implemented them in a rough way. As these modifications affect the very core of pysph, it would be great to see a general option for Lees-Edwards BCs in pysph. Do you think, that would be of interest for other users?
Cheers, Nils

PR #279, opened 2020-08-12
Improve profiling and use count sort instead of radix sort

adityapb:


pysph/base/gpu_nnps_helper.py
```diff
@@ -23,7 +23,7 @@ def get_elwise_kernel(kernel_name, args, src, preamble=""):
         ctx, args, src,
         kernel_name, preamble=preamble
     )
-    return profile_kernel(knl, kernel_name)
+    return profile_kernel(knl, kernel_name, backend='opencl')
```
prabhuramachandran, line 26: Why is this set to explicitly only the opencl backend?  Shouldn't this depend on the specific backend, what happens when CUDA is used?

pysph/base/tree/point_tree.py
```diff
@@ -23,7 +23,7 @@ class IncompatibleTreesException(Exception):
     pass
 
 
-@named_profile('neighbor_count_prefix_sum')
+@named_profile('neighbor_count_prefix_sum', backend='opencl')
```
prabhuramachandran, line 26: Similarly here.

prabhuramachandran (commented):
Could you please clarify the points raised? Thanks!

PR #281, opened 2020-09-02
added uncorrected gradient

pawansnegi:
Added UDWIJ, UDWI, UDWJ to be used to access unaltered values when DWIJ, DWI, DWJ has been corrected by a kernel correction equation.

### DIFF
--- a/docs/source/design/equations.rst
+++ b/docs/source/design/equations.rst
@@ -330,6 +330,9 @@ following:
  - ``N_NBRS``: an integer denoting the number of neighbors for the current
    destination particle with index, ``d_idx``.
 
+In case a kernel correction is used which modifies the ``DWIJ``. The
+uncorrected version can be called using ``UDWIJ``. Similarly for ``DWI`` and
+``DWJ``, we have ``UDWI`` and ``UDWJ`` for the uncorrected gradient.
 
 .. note::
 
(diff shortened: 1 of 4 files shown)


abhinavmuta:
Hi, @psnbaba there is one problem by initializing `UDWIJ[0] = DWIJ[0]`.... If I change `DWIJ` by accessing the precomputed value of `DWIJ`, _and_ if I am using `UDWIJ` it may cause a problem. Setting `gradient(XIJ, RIJ, HIJ, UDWIJ)` is an alternate way but yes the performace is hit. However I suppose if I am going to change `DWIJ` I might as well change `UDWIJ`.

Also can you address the pep8 issues. Everything else seems alright to me.

psnbaba:
> Hi, @psnbaba there is one problem by initializing `UDWIJ[0] = DWIJ[0]`.... If I change `DWIJ` by accessing the precomputed value of `DWIJ`, _and_ if I am using `UDWIJ` it may cause a problem. Setting `gradient(XIJ, RIJ, HIJ, UDWIJ)` is an alternate way but yes the performace is hit. However, I suppose if I am going to change `DWIJ` I might as well change `UDWIJ`.
> 
> Also can you address the pep8 issues. Everything else seems alright to me.

Since UDWIJ and DWIJ are evaluated before any equation initialize, loop and post_loop function get executed, there is no chance of getting UDWIJ the corrected DWIJ.
Also Fixed pep8 issues.

PR #283, opened 2020-09-10
added linear solvers for small dimension matrices

pawansnegi:
Solver for 2X2, 3X3 and 4X4 matrix using Cramer's rule. For singular matrix case, the off-diagonal terms of A matrix are ignored. which results in Shepard interpolation.

### DIFF
--- a/pysph/sph/tests/test_linalg.py
+++ b/pysph/sph/tests/test_linalg.py
@@ -1,5 +1,7 @@
 from pysph.sph.wc.linalg import (
-    augmented_matrix, gj_solve, mat_mult, mat_vec_mult
+    augmented_matrix, gj_solve, mat_mult, mat_vec_mult, deteminant_2d,
+    deteminant_3d, deteminant_4d, replace_vector_in_matrix, linear_solver_2d,
+    linear_solver_3d, linear_solver_4d
 )
 import numpy as np
 import unittest
@@ -11,6 +13,28 @@ def gj_solve_helper(a, b, n):
     result = [0.0]*n
     is_singular = gj_solve(m, n, 1, result)
     return is_singular, result
+
+
+def deteminat_helper(A, dim):
+    det = 0.0
+    if dim == 2:
+        det = deteminant_2d(A)
+    elif dim == 3:
+        det = deteminant_3d(A)
+    elif dim == 4:
+        det = deteminant_4d(A)
+    return det
+
+
+def linear_solver_helper(A, b, dim):
+    result = [0.0]*dim
+    if dim == 2:
+        linear_solver_2d(A, b, result)
+    elif dim == 3:
+        linear_solver_3d(A, b, result)
+    elif dim == 4:
+        linear_solver_4d(A, b, result)
+    return result
 
 
 class TestLinalg(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


pysph/sph/wc/linalg.py
```diff
@@ -21,6 +21,56 @@ def dot(a=[0.0, 0.0], b=[0.0, 0.0], n=3):
...
+                   A[8], A[10], A[11],
+                   A[12], A[14], A[15]])
+    c3 = deteminant_3d([A[4], A[5], A[7],
+                   A[8], A[9], A[11],
+                   A[12], A[13], A[15]])
+    c4 = deteminant_3d([A[4], A[5], A[6],
+                   A[8], A[9], A[10],
+                   A[12], A[13], A[14]])
+    result = A[0] * c1 - A[1] * c2 + A[2] * c3 - A[3] * c4
+    return result
```
abhinavmuta, line 50: Big trouble here! If I pass `A` which is defined as `A = matrix('9')` and then intialized as did in the `kernel_corrections.py` file, with `dim = n = 2` and if I call `deteminant_2d(A, b, res)` where `A` is a 9 element 1-D array with only `1, 2, 4, 5` values of the array are initialized to the correct values and the remaining `3, 6, 7, 8, 9` values possibly undefined. `deteminant_2d` will give me the wrong value!

pysph/sph/wc/linalg.py
```diff
@@ -62,6 +112,115 @@ def mat_vec_mult(a=[1.0, 0.0], b=[1.0, 0.0], n=3, result=[0.0, 0.0]):
...
+
+    detA = A[0]*A[3] - A[1]*A[2]
+    if abs(detA) > 1e-14:
+        detA1 = b[0]*A[3] - A[1]*b[1]
+        detA2 = A[0]*b[1] - b[0]*A[2]
+        result[0] = detA1/detA
+        result[1] = detA2/detA
+    else:
+        result[0] = b[0]/A[0]
+        result[1] = b[1]/A[3]
```
abhinavmuta, line 138: Same goes here! `A[3]` or others may not be initialized properly. Also, the else part is not clear to me, why divide by the digonal elemets, what if the determinant is zero as well as the diagonal elements.

pysph/sph/wc/linalg.py
```diff
@@ -21,6 +21,56 @@ def dot(a=[0.0, 0.0], b=[0.0, 0.0], n=3):
...
+        for i in range(dim*dim):
+            if index == pos:
+                result[i] = b[next]
+                next += 1
+            else:
+                result[i] = A[i]
+
+            index += 1
+            if index == dim:
+                index = 0
```
abhinavmuta, line 68: By doing this you can save on the additional condition.
```
for i in range(dim):
    for j in range(dim):
        if j = pos:
            result[i*dim + j] = b[i]
        else:
            result[i*dim + j] = A[i*dim + j]
```

abhinavmuta (changes requested):

pysph/sph/tests/test_linalg.py
```diff
@@ -1,5 +1,7 @@
 from pysph.sph.wc.linalg import (
-    augmented_matrix, gj_solve, mat_mult, mat_vec_mult
+    augmented_matrix, gj_solve, mat_mult, mat_vec_mult, deteminant_2d,
+    deteminant_3d, deteminant_4d, replace_vector_in_matrix, linear_solver_2d,
```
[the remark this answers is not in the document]
psnbaba: done

pysph/sph/tests/test_linalg.py
```diff
@@ -1,5 +1,7 @@
...
+    linear_solver_3d, linear_solver_4d
 )
 import numpy as np
 import unittest
```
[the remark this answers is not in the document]
psnbaba: done

pysph/sph/tests/test_linalg.py
```diff
@@ -200,6 +224,67 @@ def test_singular_matrix(self):
...
+        # Given
+        mat4 = [[1., 3., 6., 5.], [4., -1., 4., 1.], [3., 5., 2., 0.], [1., 8., 2., 7.]]
+        b = [1., 4., 5., 7.]
+        # when
+        result = [0.0]*16
+        ret = replace_vector_in_matrix(np.ravel(mat4), b, 4, 4, result)
+        #the
+        self.assertAlmostEqual(ret, -1.0)
+
+    def test_2d_linear_solver(self):
```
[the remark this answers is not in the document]
psnbaba: The user has to pass the matrix appropriately, if a 2D matrix is solver using a 4D array, then the elements should be at the first four positions. Similarly for 3D and 4D.
Added condition, where if a inappropriate matrix size is passed i.e. diagonal element is zero then it return 0.0 else 1.0 with the result

Added test for this too.

pysph/sph/wc/kernel_correction.py
```diff
@@ -88,29 +88,31 @@ class GradientCorrection(Equation):
...
         res = declare('matrix(3)')
+        b = declare('matrix(3)')
         eps = 1.0e-04 * HIJ
         for i in range(n):
             for j in range(n):
-                temp[nt * i + j] = d_m_mat[9 * d_idx + 3 * i + j]
-            # Augmented part of matrix
-            temp[nt*i + n] = DWIJ[i]
+                A[n * i + j] = d_m_mat[9 * d_idx + 3 * i + j]
+            b[i] = DWIJ[i]
```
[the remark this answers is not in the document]
psnbaba: In line 109, in case of ``dim==2`` only the first 4 location if the ``A`` matrix will be initialized which will be passed to the linear_solver_2d.

pysph/sph/wc/linalg.py
psnbaba: Added a check for diagonal elements, if it is zero than the solver return 0.0 instead 1.0 with the result.

pysph/sph/wc/linalg.py
psnbaba: The else part is specific to SPH, the determinant could be zero due to lack of support in those cases the result will be shepard interpolation.

pysph/sph/wc/linalg.py
psnbaba: done, Thanks.

pysph/sph/wc/kernel_correction.py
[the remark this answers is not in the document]
abhinavmuta: Yes, I didn't notice the `n*i` in the index.

PR #294, opened 2020-12-18
Fix bugs in sod shocktube.

abhinavmuta:
- Before both solids and mirror bc are set which is incorrect.
 - Solid support is removed as mirror bc is sufficient for this problem.

### DIFF
--- a/pysph/examples/gas_dynamics/sod_shocktube.py
+++ b/pysph/examples/gas_dynamics/sod_shocktube.py
@@ -4,7 +4,6 @@ from pysph.examples.gas_dynamics.shocktube_setup import ShockTubeSetup
 from pysph.sph.scheme import ADKEScheme, GasDScheme, GSPHScheme, SchemeChooser
 from pysph.sph.wc.crksph import CRKSPHScheme
 from pysph.base.nnps import DomainManager
-import numpy
 
 # Numerical constants
 dim = 1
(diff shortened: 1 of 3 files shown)


pysph/examples/gas_dynamics/sod_shocktube.py
```diff
@@ -50,41 +49,52 @@ def consume_user_options(self):
...
-            beta=2.0, update_alpha1=True, update_alpha2=True
+            fluids=['fluid'], solids=[], dim=dim, gamma=gamma,
+            kernel_factor=None, alpha1=1.0, alpha2=0.1,
+            beta=2.0, update_alpha1=True, update_alpha2=True,
         )
         gsph = GSPHScheme(
-            fluids=['fluid'], solids=['boundary'], dim=dim, gamma=gamma,
-            kernel_factor=1.0,
+            fluids=['fluid'], solids=[], dim=dim, gamma=gamma,
+            kernel_factor=None,
```
psnbaba, line 90: Setting this to None will effect values of UpdateSmoothingLengthFromVolume equation.

pysph/examples/gas_dynamics/sod_shocktube.py
```diff
@@ -50,41 +49,52 @@ def consume_user_options(self):
...
-            kernel_factor=1.2, alpha1=1.0, alpha2=0.1,
-            beta=2.0, update_alpha1=True, update_alpha2=True
+            fluids=['fluid'], solids=[], dim=dim, gamma=gamma,
+            kernel_factor=None, alpha1=1.0, alpha2=0.1,
+            beta=2.0, update_alpha1=True, update_alpha2=True,
         )
         gsph = GSPHScheme(
-            fluids=['fluid'], solids=['boundary'], dim=dim, gamma=gamma,
-            kernel_factor=1.0,
+            fluids=['fluid'], solids=[], dim=dim, gamma=gamma,
```
psnbaba, line 89: setting `solids=[]` makes empty groups for example see https://github.com/pypr/pysph/blob/7e0b5d5eb8a925943623210799a9cec436e142d9/pysph/sph/scheme.py#L1323

pysph/examples/gas_dynamics/sod_shocktube.py
abhinavmuta: This value should be same as the smoothing length factor hdx. I am setting it to hdx in configure_scheme.

pysph/examples/gas_dynamics/sod_shocktube.py
abhinavmuta: Thanks. I've put a condition here.

prabhuramachandran:
LGTM, thanks!

PR #307, opened 2021-02-06
Skip empty levels in stratified hash nnps

### LINKED ISSUE #237: Segmentation fault on non x86 platforms
We are working to prepare the debian package for pysph v1.0-a6 (python3 only) we found that there are Segmentation fault on several non-x86 platforms:

```
pysph/base/tests/test_nnps.py .......ssssssssss......................... [ 25%]
...Segmentation fault
```

See also [1] and [2].

[1] https://buildd.debian.org/status/fetch.php?pkg=pysph&arch=arm64&ver=1.0%7Ea6-3&stamp=1567840824&raw=0
[2] https://buildd.debian.org/status/package.php?p=pysph

adityapb:
Fix a division by 0 bug in stratified hash nnps. Fixes #237

### DIFF
--- a/pysph/base/stratified_hash_nnps.pyx
+++ b/pysph/base/stratified_hash_nnps.pyx
@@ -162,10 +162,16 @@ cdef class StratifiedHashNNPS(NNPS):
         cdef HashTable* hash_level = NULL
         cdef HashEntry* candidate_cell = NULL
 
-        for i from 0<=i<self.num_levels:
+        cdef double hmax_level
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Thanks!  The test error is unrelated, I was wondering if you could also add a small test case for this so there isn't a regression?

prabhuramachandran:
Merging on priority, we need to fix the shippable issue in a separate PR.

PR #308, opened 2021-03-14
BUG: update_nnps called too many times from group.

prabhuramachandran:
This happens when one sets `update_nnps=True` in a Group with multiple
destinations.  Instead of calling update_nnps once at the end, we were
calling it one for each destination which is pointless and wasteful.
This is now tested and fixed on CPU and GPU.


prabhuramachandran:
Ignoring shippable for now, will fix in a later PR.

PR #319, opened 2021-10-14
Parallelized the build step of Octree

MiloniAtal:


prabhuramachandran:
While the switch is fine, it will mean that the tests never really test the new code since the number of particles is small.  We need to also make sure that the tests run with more threads as well.

pysph/base/octree.pyx
```diff
@@ -456,7 +683,7 @@ cdef class Octree:
 
         """
         cdef NNPSParticleArrayWrapper pa_wrapper = NNPSParticleArrayWrapper(pa)
-        return self.c_build_tree(pa_wrapper)
+        return self.c_build_tree(pa_wrapper, threshold = 40000)
```
prabhuramachandran, line 686: This should be `threshold` and not `threshold=40000` otherwise setting the threshold has no effect.

prabhuramachandran (changes requested):
Apart from the one small change, this looks good.  Thanks for all the test cases!

prabhuramachandran (approved):

PR #321, opened 2021-11-01
📝 Fix for L10N sphinx document

tkoyama010:
This change is same with enthought/mayavi#777 .


prabhuramachandran:
Thanks! Merging.

PR #331, opened 2022-02-13
Corrected Surface tension models 

amalss18:
Fixed the implementations of surface tension models from the following papers:

1) M. Shadloo, M. Yildiz, "Numerical modelling of Kelvin-Helmholtz
   instability using smoothed particle hydrodynamics", IJNME, 2011
2) Adami et al. "A new surface-tension formulation for multi-phase SPH
   using a reproducing divergence approximation", JCP 2010
3) X.Y.Hu, N.A. Adams. "A multi-phase SPH method for macroscopic and
   mesoscopic flows", JCP 2006

### DIFF
--- a/pysph/sph/surface_tension.py
+++ b/pysph/sph/surface_tension.py
@@ -147,7 +147,8 @@ class MomentumEquationPressureGradientHuAdams(Equation):
         self.gx = gx
         self.gy = gy
         self.gz = gz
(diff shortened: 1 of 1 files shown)


prabhuramachandran (changes requested):
In addition to the minor changes I am asking, I notice that you have a bunch of merge commits, can you instead rebase your commits on top of current master and update your PR?

prabhuramachandran (approved):

PR #337, opened 2022-02-15
doc(install): Tip on installation of zoltan from AUR

nauaneed:
Reason to include: The zoltan AUR package seems to be actively maintained. So, it is likely to not remain broken if there are changes like if the url is updated, at least for arch users.


prabhuramachandran:
Thanks! Merged.

PR #338, opened 2022-02-15
Improved Parallel Algorithm for Octree and Compressed Octree 

MiloniAtal:
One thread launch parallel algorithm for Octree and Compressed Octree. Also fixed memory leak in the algorithm.

### DIFF
--- a/pysph/base/octree.pyx
+++ b/pysph/base/octree.pyx
@@ -11,6 +11,7 @@ from cython.operator cimport dereference as deref, preincrement as inc
 import numpy as np
 cimport numpy as np
 from cython.parallel import parallel, prange, threadid
+
 # EPS_MAX is maximum value of eps in tree building
 DEF EPS_MAX = 1e-3
 
@@ -20,7 +21,21 @@ IF UNAME_SYSNAME == "Windows":
     cdef inline double fmax(double x, double y) nogil:
         return x if x > y else y
 
-ctypedef cOctreeNode* node_ptr   
+ctypedef cOctreeNode* node_ptr
+ctypedef double* dbl_ptr
+
+cdef extern from *:
+    """
+    #define START_OMP_PARALLEL_PRAGMA() _Pragma("omp parallel") {
+    #define END_OMP_PRAGMA() }
+    #define START_OMP_SINGLE_PRAGMA() _Pragma("omp single") {
+    #define START_OMP_BARRIER_PRAGMA() _Pragma("omp barrier") {
+    """
+    void START_OMP_PARALLEL_PRAGMA() nogil
+    void END_OMP_PRAGMA() nogil
+    void START_OMP_SINGLE_PRAGMA() nogil
+    void START_OMP_BARRIER_PRAGMA() nogil
+
 ########################################################################
 
 cdef class OctreeNode:
@@ -384,7 +399,7 @@ cdef class Octree:
             node.is_leaf = True
             return 1
 
-        cdef u_int* p_indices = <u_int*> malloc(n*sizeof(u_int))
+        cdef u_int* p_indices = self.pids
         cdef vector[u_int]* child_indices = new vector[u_int](n)
         cdef vector[vector[int]] cumulative_map = vector[vector[int]](num_threads)
         cdef vector[vector[double]] threads_hmax = vector[vector[double]](num_threads)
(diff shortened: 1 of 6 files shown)


pysph/base/octree.pyx
```diff
@@ -450,127 +465,180 @@ cdef class Octree:
...
-                        continue
-                    next_level_nodes.push_back(node)
-
-        del level_nodes
+            while(count[num_threads] > 0):
+                for i from 0<=i<8:
+                    new_indices[i] = vector[u_int]()
+                children = 0
+
+                for n in xrange(old_nodes[tid].size()):
```
prabhuramachandran, line 537: Why are you using this `n in xrange`, instead why not use the `for i from ...` form or even just for i in range(...). I am not sure if this will be optimal?

pysph/base/octree.pyx
```diff
@@ -876,18 +945,357 @@ cdef class CompressedOctree(Octree):
...
+
+        if (n < self.leaf_max_particles):
+            for i from 0<=i<n:
+                self.pids[i] = i
+            node.is_leaf = True
+            return 1
+
+        cdef u_int* p_indices = self.pids
+        cdef vector[u_int]* child_indices = new vector[u_int](n)
+        cdef vector[vector[int]] cumulative_map = vector[vector[int]](num_threads)
```
prabhuramachandran, line 983: Add a comment here to explain what this is going to store, i.e. cumulative_map[thread_id][octant_id] or so.

pysph/base/octree.pyx
```diff
@@ -876,18 +945,357 @@ cdef class CompressedOctree(Octree):
...
+        cdef vector[vector[double]] threads_hmax = vector[vector[double]](num_threads)
+
+        cdef vector[vector[double]] threads_xmin = vector[vector[double]](num_threads)
```
prabhuramachandran, line 986: Does it make sense to put all these into a single array with 7 elements, i.e. xmin, xmax, ymin, ymax, zmin, zmax, hmax? If this makes it harder to read then leave it.

pysph/base/octree_nnps.pyx
```diff
@@ -225,7 +229,7 @@ cdef class CompressedOctreeNNPS(OctreeNNPS):
     cpdef _refresh(self):
         cdef int i
         for i from 0<=i<self.narrays:
-            (<CompressedOctree>self.tree[i]).c_build_tree(self.pa_wrappers[i])
+            (<CompressedOctree>self.tree[i]).c_build_tree(self.pa_wrappers[i], self.tree[i].test_parallel)
```
prabhuramachandran, line 232: Can you wrap this to 78 cols?

pysph/base/octree.pyx
```diff
@@ -621,23 +689,23 @@ cdef class Octree:
 
         cdef int num_threads = get_number_of_threads()
 
-        #Use the parallel method
-        if ( (num_threads > 4 and num_particles > 40000) or test_parallel ):
+        # Use the parallel method
+        if ( (num_threads > 1) or test_parallel ):
             self.method = 1
```
prabhuramachandran, line 694: Is this self.method hack needed anymore given that there is just a single test_parallel case.

prabhuramachandran (changes requested):
Please take a look and see if any of these can be incorporated. Thanks!

pysph/base/octree_nnps.pxd
```diff
@@ -21,7 +21,7 @@ cdef class OctreeNNPS(NNPS):
     ##########################################################################
     # Data Attributes
     ##########################################################################
-    cdef public list tree
+    cdef list tree
```
prabhuramachandran, line 24: Just curious, why have you made this non-public?

pysph/base/octree.pyx
```diff
@@ -875,19 +944,373 @@ cdef class CompressedOctree(Octree):
...
+            count[d] = count[d] + 1
+            old_nodes[d].push_back(deref(level_nodes)[i])
+            d = (d + 1)%num_threads
+        count[num_threads] = num_nodes
+        count[num_threads + 1] = 2
+
+        del level_nodes
+
+        with nogil, parallel():
+            xmin_current = <double *> malloc(3 * sizeof(double))
```
prabhuramachandran, line 1160: Do you need to malloc this? Can you not allocate this on the stack with a `double[3]`?  I ask because I don't see a matching free in this PR or in the sources.  For every malloc there should be a matching free. Similarly for every "new" a corresponding "del".  I think there may still lurk some issues like this.

prabhuramachandran (changes requested):
A few minor points but one serious issue with no matching free/dels, can you just make sure or add suitable comments to make sure there is no small memory leak.

pysph/base/octree_nnps.pxd
MiloniAtal: I had actually had to enable this for self.method hack. So, I re-turned it to non-public state now.

prabhuramachandran (approved):

PR #344, opened 2022-03-05
ci(gh actions): add caching

nauaneed:
- cache auto-generated code in ~/.pysph and ~/.compyle
- refresh cache key monthly
- new monthly cron job


prabhuramachandran:
Awesome, thank you!

PR #353, opened 2022-06-03
Add name keyword to Group class useful while profiling.

abhinavmuta:


pysph/sph/equation.py
```diff
@@ -530,6 +543,9 @@ def __init__(self, equations, real=True, update_nnps=False, iterate=False,
         self.condition = condition
         self.start_idx = start_idx
         self.stop_idx = stop_idx
+        self.name = 'Group_%d' %group_counter.__next__()
```
prabhuramachandran, line 546: Use `% next(group_counter)`

pysph/sph/equation.py
```diff
@@ -549,7 +565,7 @@ def __init__(self, equations, real=True, update_nnps=False, iterate=False,
     # Non-public interface.
     ##########################################################################
     def __repr__(self):
-        cls = self.__class__.__name__
+        cls = self.name
```
prabhuramachandran, line 568: No, this should not be changed as a repr should also show how one can instantiate the group. Leave the `cls`, just make sure that name=xxx shows in the constructor arguments to the Group.

pysph/sph/acceleration_eval_cython.mako
```diff
@@ -37,17 +40,21 @@ ${indent(all_eqs.get_py_initialize_code(), 0)}
 #######################################################################
```
prabhuramachandran, line 40: You could also add a line for the `py_initialize`?  Although this will need to be done in the helper and not in the mako.

pysph/sph/acceleration_eval_cython.mako
```diff
@@ -135,6 +150,7 @@ with profile_ctx("nnps.update"):
 % if group.post:
 ${indent(helper.get_post_call(group), 0)}
```
prabhuramachandran, line 151: You have not covered the Group.post here.

prabhuramachandran (changes requested):
Thanks! Please make these changes. Please also fix the PEP8 errors.

prabhuramachandran:
@abhinavmuta -- the test error is because the level changes due to the addition of the global profile. Please also fix this. In addition, for the pure Python calls that are profiled, use the `with profile_ctx(...)` form as done for the update_nnps.

pysph/sph/equation.py
abhinavmuta: Thanks, Fixed.

pysph/sph/acceleration_eval_cython.mako
abhinavmuta: I've added the `with profile_ctx(...)` in the `equation.py` file.

pysph/sph/acceleration_eval_cython.mako
abhinavmuta: I've added the `with profile_ctx(...)` in the cython helper file as the pre and post are pure python calls.

pysph/sph/equation.py
abhinavmuta: Thanks, fixed.

prabhuramachandran (approved):

PR #356, opened 2022-08-05
Generate XDMF

nauaneed:
Instead of `dump_vtk` which reads every npz or hdf5 file and creates a vtk file, XDMF files which just reference the heavy data stored in hdf5 files can be generated using the new `generate_xdmf` sub-command.


prabhuramachandran:
Please add a simple test case to check the generated xdmf, also I suggest using `dump_xdmf` rather than `generate_xdmf` to be consistent with `dump_vtk`. Also the vectorize velocity could be the default -- is there a case where it is not useful, if not maybe an option to not do vectorization of velocity.

PR #357, opened 2022-08-09
Dump XDMF

nauaneed:
Improved https://github.com/pypr/pysph/pull/356

1. `generate_xdmf` is now `dump_xdmf`.
2. vectorize velocity is now default.
3. added a test case to check generated xdmf.
4. hdf5 files are are referred by relative paths in xdmf by default.

### DIFF
--- a/pysph/tools/tests/test_dump_xdmf.py
+++ b/pysph/tools/tests/test_dump_xdmf.py
@@ -1,0 +1,73 @@
+import shutil
+import unittest
+from pathlib import Path
+from tempfile import mkdtemp
+
+import numpy as np
+from pytest import importorskip
+
+from pysph.base.utils import get_particle_array
+from pysph.solver.output import dump
+from pysph.tools.dump_xdmf import main as dump_xdmf
+
+
+class TestDumpXDMF(unittest.TestCase):
+
+    def test_dump_xdmf(self, npoints=10, random_seed=0):
+        vtk = importorskip('vtk')
+        np.random.seed(random_seed)
+        tmp_dir = mkdtemp()
+        hdf5file = str(Path(tmp_dir) / 'test.hdf5')
+
+        # Assign rho as random data, make a particle array and dump it as hdf5.
+        self.rho = np.random.rand(npoints)
+        pa = get_particle_array(name='fluid',
+                                rho=self.rho,
+                                x=np.arange(npoints),
+                                y=np.zeros(npoints),
+                                z=np.zeros(npoints))
+        dump(hdf5file, [pa], {})
+
+        try:
+            # Generate XDMF for dumped hdf5 file.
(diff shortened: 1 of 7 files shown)


nauaneed:
Not sure why this fails on windows. I will try to run this test on windows manually to find out.

pysph/tools/tests/test_dump_xdmf.py
```diff
@@ -0,0 +1,60 @@
+import shutil
+import unittest
+from pathlib import Path
+from tempfile import mkdtemp
+
+import numpy as np
+import vtk
```
prabhuramachandran, line 7: OK this is a very hard and difficult dependency and I would rather keep this as an optional one. Can you use a `pytest.importorskip` to avoid baking this into a hard dependency please?

requirements-test.txt
```diff
@@ -1,3 +1,4 @@
 pytest>=3.0
 mock>=1.0
 h5py
+vtk
```
prabhuramachandran, line 4: Is there a way to test without using vtk?

setup.py
```diff
@@ -684,7 +684,7 @@ def setup_package():
         'numpy', 'mako', 'cyarray', 'compyle>=0.8', 'Cython>=0.20',
         'setuptools>=42.0.0', 'pytools', 'Beaker'
     ]
-    tests_require = ['pytest>=3.0', 'h5py']
+    tests_require = ['pytest>=3.0', 'h5py', 'vtk']
```
prabhuramachandran, line 687: Is there a way to test this without using VTK?

prabhuramachandran (changes requested):
This is great but I am not happy about adding VTK in as a test dependency.

pysph/tools/tests/test_dump_xdmf.py
nauaneed: Yes, I will do this.

setup.py
nauaneed: Yes, I would have to write my own xdmf reader but that would be really complicated. 
Also, if my xdmf reader reads it but somehow vtk's xdmf reader cannot, then I think it wouldn't be really useful.

prabhuramachandran: Is there a pure Python reader that you can use? I think `meshio` will do the job pretty well, and I think it may already be an optional requirement.

nauaneed: I have tried `meshio` earlier because that would have been easier. I went with `vtk` since `meshio` had limited xdmf support. Other than that, xdmf.org has [something](https://gitlab.kitware.com/xdmf/xdmf) but that is also not pure python.

prabhuramachandran (approved):

prabhuramachandran:
Given that this is not going to be tested in the CI, should we just add the vtk requirement back in tests_require and in `requirements-test.txt`?

PR #358, opened 2022-08-09
New magma2 scheme

nauaneed:


pysph/sph/gas_dynamics/magma2.py
```diff
@@ -0,0 +1,1771 @@
...
+    while i <= j:
+        while arr[i] < pivot:
+            i += 1
+        while arr[j] > pivot:
+            j -= 1
+
+        if i <= j:
+            arr[i], arr[j] = arr[j], arr[i]
+            i, j = i + 1, j - 1
+    quicksort(arr, fst, j)
```
prabhuramachandran, line 1770: This will unfortunately not work on a GPU but that is OK for now.

prabhuramachandran (commented):
This looks good, am not sure if you intend to use this on a GPU or not but that is OK for now.

PR #360, opened 2022-10-08
ci(actions): fix tests failing on windows

nauaneed:
- fixed by using `python setup.py develop` instead of `pip install -e .`

- cause ? : https://github.com/pypa/pip/issues/11467


prabhuramachandran:
Thanks a ton! Awesome.

PR #368, opened 2023-01-22
SPLASH Interpolator

nauaneed:
An example where this interpolator performs better than the existing interpolator:

Attempting to interpolate the central region marked by the square from the following solution onto a 1000x1000 grid.
riemann_2d

#### Existing `pysph.tools.interpolator.Interpolator`
pysph-existing

#### New `pysph.tools.interpolator.SplashInterpolator`
pysph-splash-new

#### [SPLASH](https://users.monash.edu.au/~dprice/splash/)
original-splash


prabhuramachandran:
Cool, please note that it may be easier to just add this to the existing interpolator, as an option since you can pass in the `method` to the constructor.

PR #371, opened 2023-02-21
Reuse interpolators instead of creating afresh.

prabhuramachandran:
Earlier this was not a problem but a recent PR has introduced group names, so each time the interpolator is constructed, it creates new Equation groups which have their own names and each name is unique, so this means the compilation is performed on each data file being processed. I have changed this for now to reuse the interpolator while post-processing.


abhinavmuta (approved):
Looks good to me.

PR #376, opened 2023-04-04
Added playback buttons to viewer

nauaneed:
Screenshot_2023-04-04-20-10-11_1920x1080

The unicode characters show up in windows but they are blue.


prabhuramachandran:
Thanks! LGTM.

PR #390, opened 2023-11-28
Fix build and tests

nauaneed:


prabhuramachandran (approved):
Looks good to me for now. A more careful fix that works on cython 3.0 would be useful.

PR #391, opened 2023-12-03
Compatibility with cython 3.0

avalentino:


prabhuramachandran:
I think there are other issues with Cython 3.0 which significantly affect performance of pysph. We will debug these later so PR #390 should address the build failures for now, although this precludes the possibility of installing on 3.12 yet.

prabhuramachandran:
hank you. Closing as this is resolved by https://github.com/pypr/pysph/pull/400

PR #400, opened 2024-08-18
Bump cython, fixes for tests, and fixes for deprecation warnings

nauaneed:


nauaneed:
Green, yay!

pysph/base/c_kernels.pyx.mako
```diff
@@ -12,6 +12,7 @@ CLASSES = (
     WendlandQuinticC4, WendlandQuinticC6, WendlandQuinticC2_1D,
     WendlandQuinticC4_1D, WendlandQuinticC6_1D
 )
+get_config().use_openmp = True;
```
prabhuramachandran, line 15: Why is this being done? Also, if you are changing it, should you not reset it to the original state? So my suggestion is to use the with_config or get the original value and restore it.

prabhuramachandran (changes requested):
Barring the one issue, the rest looks great. Thanks a lot for fixing this.

prabhuramachandran (approved):
LGTM, thanks!

PR #406, opened 2025-04-18
Fixes and enhancements

nauaneed:
Fixes and enhancements to:
- solver
- viewer
- interpolator
- dump_xdmf

### DIFF
--- a/pysph/tools/interpolator.py
+++ b/pysph/tools/interpolator.py
@@ -1,6 +1,6 @@
 # Standard library imports
 from functools import reduce
-
+interpolator_methods = ['shepard', 'sph', 'order1', 'splash', 'splash_norm']
 # Library imports.
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


pysph/tools/interpolator.py
```diff
@@ -1,6 +1,6 @@
 # Standard library imports
 from functools import reduce
-
+interpolator_methods = ['shepard', 'sph', 'order1', 'splash', 'splash_norm']
```
prabhuramachandran, line 3: This could be a class attribute on the `Interpolator` class, say `Interpolator.METHODS` or something like that?

prabhuramachandran (changes requested):
LGTM barring the one small suggestion.

pysph/tools/interpolator.py
nauaneed: Yes!

prabhuramachandran (approved):
LGTM!

